PR #2, opened 2015-07-22
Use `::libc::c_char` instead of `i8`

akiss77:
Some platforms have unsigned chars, there using `i8` causes
compilation errors.


pcwalton:
Conflicts now with my Linux rewrite; could you redo this? Thanks :)

akiss77:
Redone :)

PR #7, opened 2015-09-22
Add in-process mpsc-channels based implementation

vvuk:
This PR adds an in-process-only implementation on top of mpsc channels.  It uses this as the default on windows until we finish a real windows implementation.

### DIFF
--- a/platform/inprocess/mod.rs
+++ b/platform/inprocess/mod.rs
@@ -21,30 +21,26 @@ use uuid::Uuid;
 
 struct SenderRecord {
     sender: MpscSender,
-    lock: Mutex<bool>,
-    condvar: Condvar,
+    conn_sender: mpsc::Sender<bool>,
+    conn_receiver: Mutex<mpsc::Receiver<bool>>,
 }
 
 impl SenderRecord {
     fn new(sender: MpscSender) -> SenderRecord {
+        let (tx, rx) = mpsc::channel::<bool>();
         SenderRecord {
             sender: sender,
-            lock: Mutex::new(false),
-            condvar: Condvar::new(),
+            conn_sender: tx,
+            conn_receiver: Mutex::new(rx),
         }
     }
 
-    fn wait_for_acquire(&self) {
-        let mut started = self.lock.lock().unwrap();
-        while !*started {
-            started = self.condvar.wait(started).unwrap();
-        }
+    fn accept(&self) {
+        self.conn_receiver.lock().unwrap().recv().unwrap();
     }
 
-    fn signal_acquisition(&self) {
-        let mut notified = self.lock.lock().unwrap();
-        *notified = true;
-        self.condvar.notify_one();
+    fn connect(&self) {
+        self.conn_sender.send(true).unwrap();
     }
 }
 
@@ -116,7 +112,7 @@ impl MpscSender {
 
     pub fn connect(name: String) -> Result<MpscSender,MpscError> {
         let record = ONE_SHOT_SENDERS.lock().unwrap().remove(&name).unwrap();
-        record.signal_acquisition();
+        record.connect();
         Ok(record.sender)
     }
 
@@ -242,7 +238,7 @@ impl MpscOneShotServer {
                                     Vec<OpaqueMpscChannel>,
                                     Vec<MpscSharedMemory>),MpscError>
     {
-        ONE_SHOT_SENDERS.lock().unwrap().get(&self.name).unwrap().wait_for_acquire();
+        ONE_SHOT_SENDERS.lock().unwrap().get(&self.name).unwrap().accept();
         let receiver = self.receiver.borrow_mut().take().unwrap();
         let (data, channels, shmems) = receiver.recv().unwrap();
         Ok((receiver, data, channels, shmems))
--- a/platform/inprocess/mod.rs
+++ b/platform/inprocess/mod.rs
@@ -1,0 +1,353 @@
+// Copyright 2015 The Servo Project Developers. See the COPYRIGHT
+// file at the top-level directory of this distribution.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+use std::sync::mpsc;
+use std::sync::{Arc, Mutex, Condvar};
+use std::collections::hash_map::HashMap;
+use std::cell::{RefCell};
+use std::slice;
+use std::fmt::{self, Debug, Formatter};
+use std::cmp::{PartialEq};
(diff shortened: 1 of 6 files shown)


platform/inprocess/mod.rs
```diff
@@ -0,0 +1,359 @@
...
+
+            let id = select.wait();
+            //println!("select id: {} length: {}", id, receivers.len());
+
+            for (index,h) in handles.iter().enumerate() {
+                //println!("index {} id {}", index, id);
+                if h.id() == id {
+                    r_index = index;
+                    r_id = self.receiver_ids[index] as i64;
+                    //println!("r_id {}", r_index);
```
pcwalton: nit: remove these commented out printlns

platform/inprocess/mod.rs
```diff
@@ -0,0 +1,359 @@
...
+
+    pub fn send(&self,
+                data: &[u8],
+                ports: Vec<MpscChannel>,
+                shared_memory_regions: Vec<MpscSharedMemory>)
+                -> Result<(),MpscError>
+    {
+        match self.sender.borrow().send(MpscChannelMessage(data.to_vec(), ports, shared_memory_regions)) {
+            Err(_) => { Err(MpscError::ChannelClosedError) },
+            Ok(_) => { Ok(()) },
```
pcwalton: nit: no need for the braces here

platform/inprocess/mod.rs
```diff
@@ -0,0 +1,359 @@
...
+    pub fn new() -> Result<MpscReceiverSet,MpscError> {
+        Ok(MpscReceiverSet {
+            last_index: 0,
+            receiver_ids: vec![],
+            receivers: vec![],
+        })
+    }
+
+    pub fn add(&mut self, receiver: MpscReceiver) -> Result<i64,MpscError> {
+        self.last_index = self.last_index + 1;
```
pcwalton: nit: `self.last_index += 1;` is shorter

platform/inprocess/mod.rs
```diff
@@ -0,0 +1,359 @@
...
+            mem::replace(&mut *r.receiver.borrow_mut(), mem::replace(&mut receivers[index], None));
+        }
+
+        if r_id == -1 {
+            return Err(MpscError::UnknownError);
+        }
+
+        let receivers = &mut self.receivers;
+        match receivers[r_index].recv() {
+            Ok((data, channels, shmems)) => return Ok(vec![MpscSelectionResult::DataReceived(r_id, data, channels, shmems)]),
```
pcwalton: nit: I don't think you need the `return` here or on any of the other branches

platform/inprocess/mod.rs
```diff
@@ -0,0 +1,359 @@
...
+            Err(err) => {
+                match err {
+                    MpscError::ChannelClosedError => {
```
pcwalton: Could just be folded into the match above: `Err(MpscError::ChannelClosedError) => { ... } Err(_) => { ... }`

platform/inprocess/mod.rs
```diff
@@ -0,0 +1,359 @@
...
+    }
+
+    fn wait_for_acquire(&self) -> () {
+        let mut started = self.lock.lock().unwrap();
+        while !*started {
+            started = self.condvar.wait(started).unwrap();
+        }
+    }
+
+    fn signal_acquisition(&self) -> () {
```
pcwalton: nit: no need for `-> ()`

platform/inprocess/mod.rs
```diff
@@ -0,0 +1,359 @@
...
+        let mut notified = self.lock.lock().unwrap();
+        *notified = true;
```
pcwalton: Is there a reason why this lock/condvar code can't be just an MPSC `Mutex<Receiver<()>>`/`Sender<()>` pair? That is, clients just lock the receiver and wait. This code isn't performance critical and I find channels easier to think about than condvars (maybe it's just from writing a lot of Servo code, though).

platform/inprocess/mod.rs
```diff
@@ -0,0 +1,359 @@
...
+    pub fn consume(&self) -> MpscReceiver {
+        let receiver = self.receiver.borrow_mut().take();
+        MpscReceiver::new(receiver.unwrap())
+    }
+
+    pub fn recv(&self) -> Result<(Vec<u8>, Vec<OpaqueMpscChannel>, Vec<MpscSharedMemory>),MpscError> {
+        let r = self.receiver.borrow();
+        match r.as_ref().unwrap().recv() {
+            Ok(MpscChannelMessage(d,c,s)) => Ok((d,
+                                                 c.into_iter().map(|x| OpaqueMpscChannel::new(x)).collect(),
```
pcwalton: Does `c.into_iter().map(OpaqueMpscChannel::new).collect()` work?

platform/inprocess/mod.rs
```diff
@@ -0,0 +1,359 @@
...
+
+        {
+            let select = mpsc::Select::new();
+            // we *must* allocate exact capacity for this, because the Handles *can't move*
+            let mut handles: Vec<mpsc::Handle<MpscChannelMessage>> = Vec::with_capacity(self.receivers.len());
+        
+            for r in &self.receivers {
+                let inner_r = mem::replace(&mut *r.receiver.borrow_mut(), None);
+                receivers.push(inner_r);
+            }
```
pcwalton: Instead of swapping out each receiver individually, is it possible to do something like `let receivers = mem::replace(&mut self.receivers, Vec::new()).into_iter().map(Some).collect()`?

vvuk: I don't think so, at least not something that simple... the loop I have is reaching replacing the MpscReceiver's RefCell<Option<>> member value's with None, and pulling out the inner mpsc::Receiver.  It's not just replacing the self.receivers array values themselves (they don't change).

I could probably write it as a map that does a mem::replace, but I don't know that it would be any clearer.

platform/inprocess/mod.rs
vvuk: In theory I'd need the sender to also block until the receiver actually finishes receiving.  But maybe that's not actually necessary in this case, since the actual connection setup can't really fail...

vvuk:
Okay, converted to using a mpsc::channel() instead of a mutex/condvar (and renamed ONE_SHOT_SENDERS to ONE_SHOT_SERVERS, derp).  Though now I'm actually confused as to how this is working at all -- OneShotServer::accept() does a get on the ONE_SHOT_SERVERS hashmap, and then calls accept() which calls recv(), which should block.

MpscSender::connect() does a remove on the ONE_SHOT_SERVERS hashmap and calls send() to poke the server.  But the server part should be blocked in recv() inside accept()... so who owns the ServerRecord?

pcwalton:
I'm confused—are you asking who owns the memory for the `ServerRecord`? I don't think it matters, since all of its elements are heap allocated and atomically reference counted.

vvuk:
Yeah, I guess I'm getting a copy during the get().  This should be good to go.

PR #8, opened 2015-09-30
Bump up the size of the internal Mach buffer to the maximum.

pcwalton:
The default is 5 messages before blocking. This is too few in practice
and causes deadlocks in iframe reftests. The maximum is 1024, which
ought to be enough.

r? @larsbergstrom


larsbergstrom:
This sounds good. I believe it's even worse than 5, from what I can tell - in OSX 10.10, while the constant for the default is still 5, the internal default is reported to have changed to 2, which may be why we're seeing more of these blocking issues recently.

Moving to the default "large" value of 1024 seems like the right thing to do here.

@bors-servo r+

PR #10, opened 2015-09-30
Guard against null pointers (which can exist if the size is zero) in

pcwalton:
`MachSharedMemory::clone()` and `MachSharedMemory::deref()`.

Part of the fix for servo/servo#7422.


larsbergstrom:
r+

PR #15, opened 2015-11-22
Android portability: use socklen_t instead of c_uint in some places.

SimonSapin:
CC https://github.com/servo/servo/pull/8446


SimonSapin:
It doesn’t look like homu is set up here.

larsbergstrom:
r? @pcwalton 

(blocking rust upgrade)

PR #17, opened 2015-11-23
Return `io::Error` types instead of unit.

pcwalton:
Should help diagnose intermittent Servo test failures.

r? @larsbergstrom


larsbergstrom:
r+

PR #19, opened 2015-12-01
Switch android to use the in-process ipc bypass.

glennw:
This is a temporary workaround for some performance issues, and some android specific IPC crashes.


pcwalton:
This seems pretty unfortunate; can we make a metabug in Servo so we can get these fixed?

glennw:
Done - https://github.com/servo/servo/issues/8771

PR #21, opened 2015-12-13
Derive Debug on some of the public types

antrik:
Some of the types already have had working Debug implementations (or
derives) available in all the OS-specific backends; but for some reason
they were not exposed on the (portable) API types... Let's change that.

Note that this doesn't cover all of the public types -- some would
actually need Debug implementations to be newly introduced for all the
OS-specific backends, which would be much more involved than just
exposing the existing ones...

This is not tested on MacOS; however, the backends seem to be very symmetric -- so hopefully it should just work...

### DIFF
--- a/ipc.rs
+++ b/ipc.rs
@@ -50,6 +50,7 @@ pub fn channel<T>() -> Result<(IpcSender<T>, IpcReceiver<T>),Error>
     Ok((ipc_sender, ipc_receiver))
 }
 
+#[derive(Debug)]
 pub struct IpcReceiver<T> where T: Deserialize + Serialize {
     os_receiver: OsIpcReceiver,
     phantom: PhantomData<T>,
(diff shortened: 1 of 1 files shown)


Wafflespeanut:
huh

    platform/linux/mod.rs:23:5: 23:41 error: unresolved import `std::slice::bytes::MutableByteVector`. Could not find `bytes` in `std::slice` [E0432]

    platform/linux/mod.rs:23 use std::slice::bytes::MutableByteVector;

                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

antrik:
Hrm. The auto-builder appears to be using a different Rust snapshot than Servo. I actually got similar errors (unrelated to my changes I'm quite sure) when compiling locally with a different snapshot; but it builds fine when using the same compiler as Servo...

antrik:
It builds again (both master and my branch) with current nightly.

antrik:
[knock knock] Anyone there?...

frewsxcv:
Aster doesn't compile on nightly: https://github.com/serde-rs/aster/issues/58

PR #22, opened 2015-12-26
Fix: Provide dummy Debug impls for platform::inprocess::{MpscSender, …

antrik:
…MpscReceiver}

Fixes a regression introduced by
410a6140fd75c070cad541611686c8e30c3965c3 (
https://github.com/servo/ipc-channel/pull/21 ), which broke compilation
for targets using the inprocess backend.


antrik:
A nicer fix will be possible once https://github.com/rust-lang/rust/issues/30563 is taken care of.

PR #24, opened 2015-12-30
Bump version to 0.1.1

nox:


nox:
@metajack Should that be bumped to 1.0 directly?

Ms2ger:
I think we should

nox:
And we should clean `Cargo.toml` too, then:

```
unused manifest key: dependencies.serde.opt_level
```

Ms2ger:
We're using this from git still, so no real reason to keep this PR until we move to crates.io.

PR #25, opened 2016-01-04
Linux: Fix corruption of big transfers on 32 bit systems

antrik:
The wrong size calculation magically worked on 64 bit systems, because
2 * sizeof u32 is the same as sizeof usize there -- but not on 32 bit
systems, where this dropped 4 bytes of each full fragment.

With this, Servo works again for me. (And we are down to 4 failing ipc-channel tests on my system...)


pcwalton:
Oh my goodness. Thank you for digging into this!

PR #27, opened 2016-01-08
Linux: Don't pass uninitialised memory as pointers to libc::connect()

antrik:
I don't know whether there are constellations where taking a reference
to a binding containing mem::uninitialized() actually creates a proper
NULL pointer -- but it certainly doesn't on my system when using the
rustc snapshot from Servo.

Either way, creating explicit NULL pointers seems way more correct.

This fixes the other four testsuite failures, as well as servo -M

### DIFF
--- a/platform/linux/mod.rs
+++ b/platform/linux/mod.rs
@@ -474,9 +474,9 @@ impl UnixOneShotServer {
                                    Vec<OpaqueUnixChannel>,
                                    Vec<UnixSharedMemory>),UnixError> {
         unsafe {
-            let mut sockaddr = mem::uninitialized();
-            let mut sockaddr_len = mem::uninitialized();
-            let client_fd = libc::accept(self.fd, &mut sockaddr, &mut sockaddr_len);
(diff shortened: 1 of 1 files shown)


Ms2ger:
What testsuite failures are you referring to?

antrik:
All the tests using IpcOneShotServer are failing on my system without this fix -- including platform::test::server, platform::test::cross_process, platform::test::cross_process_sender_transfer, and test::cross_process_embedded_senders. (And servo -m accordingly also panics on IpcOneShotServer.)

Ms2ger:
Would be nice if we could test this in automation. This looks correct from the manpage, though.

@bors-servo r+

Ms2ger:
And that's https://github.com/rust-lang/rust/issues/30713

ecoal95:
Heh, now aster/quasi_codegen is broken, great

KiChjang:
serde-rs/quasi#28

ecoal95:
Dang it, not published yet

Ms2ger:
\o/

Issue #11, opened 2015-10-08
Note suggesting to use mkstemp or mkdtemp instead of mktemp


connorimes:
FYI, I get the following note when building on ARM when everything is being linked at the end.  Don't recall if I saw it elsewhere.

<pre>
note: /local/servo/components/servo/target/debug/deps/libipc_channel-948af358b3c092a6.rlib(ipc_channel-948af358b3c092a6.0.o): In function `platform::linux::_$LT$impl$GT$::new::he43c9a17ef03c572fbb':
/home/username/.cargo/git/checkouts/ipc-channel-62de6a747162919c/master/platform/linux/mod.rs:433: warning: the use of `mktemp' is dangerous, better use `mkstemp' or `mkdtemp'
</pre>

craftytrickster:
I am getting this message when trying to build servo as well on Linux (Ubuntu 14.04), with 
rust nightly date of - 2015-12-09
```
note: /usr/bin/ld.gold: error: cannot find -lEGL
/home/divad/Rust/servo/.cargo/git/checkouts/ipc-channel-3840821e54ea11a6/master/platform/linux/mod.rs:437: warning: the use of `mktemp' is dangerous, better use `mkstemp' or `mkdtemp'
collect2: error: ld returned 1 exit status

error: aborting due to previous error
Could not compile `servo`.
```

antrik:
@craftytrickster FWIW, your error was never related to the issue here, which is just a warning. (Your actual error was the `cannot find -lEGL` bit.) Nevertheless, your PR addressing the warning should have been merged... Too bad I didn't know about it before :-(

craftytrickster:
I haven't actually been able to contribute to servo in a really long time, so no worries! I'm glad you are able to resolve the issue though.

PR #28, opened 2016-01-15
Replacing unsafe mktemp with mkdtemp to avoid linux compiler errors

craftytrickster:
I am unable to compile servo on my linux laptop anymore due the following [issue](https://github.com/servo/ipc-channel/issues/11) - which is caused by the inclusion of mktemp.

Unfortuntately, using mkstemp is not a viable solution, since it actually opens a handle on the newly created file, which causes libc bind to fail. As a workaround, I thought of using mkdtemp, which safely creates a random directory, and then just adding a file there to use. I have not worked with these libc functions before, so if I am overlooking something please let me know.

### DIFF
--- a/platform/linux/mod.rs
+++ b/platform/linux/mod.rs
@@ -434,9 +434,11 @@ impl UnixOneShotServer {
             loop {
                 let path_string = CString::new(b"/tmp/rust-ipc-socket.XXXXXX" as &[u8]).unwrap();
                 path = path_string.as_bytes().iter().cloned().collect();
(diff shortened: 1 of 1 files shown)


platform/linux/mod.rs
```diff
@@ -927,4 +930,3 @@ struct pollfd {
     events: c_short,
     revents: c_short,
 }
-
```
KiChjang: Add these newline back in, please.

craftytrickster:
After getting the latest dependencies from `apt-get install`, I no longer was getting the error message. If you would like, I can close the PR.

PR #31, opened 2016-01-26
Expose more meaningful deserialization errors.

jdm:


jdm:
Let's see what Travis says here, since I don't have linux handy.

PR #38, opened 2016-02-21
Fix concurrent large transfers by using dedicated channel

antrik:
As evidenced by the newly added "concurrent_senders" test case, the
previous approach never actually worked with concurrent large transfers:
when trying to "push back" packets to the "receiver" FD of the socket
pair, they do not show up on the "receiver" FD again, but rather the
"sender" FD.

This problem can be avoided entirely by automatically creating a
dedicated channel for each large transfer, so the followup fragments can
all be received in order, without interleaving with fragments from other
senders.

The first fragment of each large transfer is still sent through the main
channel, along with an FD for the dedicated channel; the remaining
fragments are then sent through the dedicated channel.

This approach adds a slight overhead for fragmented transfers when there
is no actual concurrency -- though it is way more efficient when actual
concurrency occurs than the push-back approach would have been...

With the dedicated channel, the follow-up fragments do not really need
the fragmentation headers anymore; and the header for the initial
fragment could be simplified as well. However, to keep changes minimal,
this commit doesn't change the header yet -- this is left for later
improvements.

### DIFF
--- a/platform/linux/mod.rs
+++ b/platform/linux/mod.rs
@@ -125,7 +125,7 @@ impl UnixSender {
 
     pub fn send(&self,
                 data: &[u8],
-                channels: Vec<UnixChannel>,
+                mut channels: Vec<UnixChannel>,
                 shared_memory_regions: Vec<UnixSharedMemory>)
                 -> Result<(),UnixError> {
         let mut data_buffer = vec![0; data.len() + mem::size_of::<u32>() * 2];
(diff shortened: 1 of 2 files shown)


antrik:
According to @glennw this fixes hangs in WebRender.

platform/linux/mod.rs
```diff
@@ -137,57 +137,65 @@ impl UnixSender {
...
+                    *cmsg_padding_ptr = *DEV_NULL;
+                    cmsg_padding_ptr = cmsg_padding_ptr.offset(1);
+                }
 
-            let mut iovec = iovec {
-                iov_base: data_buffer.as_ptr() as *const c_char as *mut c_char,
-                iov_len: data_buffer.len() as size_t,
-            };
+                // Put this on the heap so address remains stable across function return.
+                let mut iovec = Box::new(iovec {
```
ecoal95: Isn't this relying on the `iovec` not being freed? (so we're either using it after free or leaking it?)

platform/linux/mod.rs
```diff
@@ -137,57 +137,65 @@ impl UnixSender {
...
-                msg_name: ptr::null_mut(),
-                msg_namelen: 0,
-                msg_iov: &mut iovec,
-                msg_iovlen: 1,
-                msg_control: cmsg_buffer as *mut c_void,
-                msg_controllen: CMSG_SPACE(cmsg_length as size_t),
-                msg_flags: 0,
+                // Be sure to always return iovec -- whether the caller uses it or not --
+                // to prevent premature deallocation!
+                (msghdr, iovec)
```
ecoal95: Whoops, forget that

platform/linux/mod.rs
ecoal95: I don't know how critical it is, but maybe you could prevent this allocation, and the hackish iovec return passing an uninitialized `&mut iovec` to the closure.

platform/linux/mod.rs
```diff
@@ -137,57 +137,65 @@ impl UnixSender {
...
-            let mut cmsg_padding_ptr =
-                (cmsg_buffer.offset(1) as *mut _ as *mut c_int).offset(fds.len() as isize);
-            let cmsg_end =
-                (cmsg_buffer as *mut _ as *mut u8).offset(CMSG_SPACE(cmsg_length as size_t) as
-                                                          isize);
-            while (cmsg_padding_ptr as *mut u8) < cmsg_end {
-                *cmsg_padding_ptr = *DEV_NULL;
-                cmsg_padding_ptr = cmsg_padding_ptr.offset(1);
-            }
+            let construct_header = |channels: &Vec<UnixChannel>, data_buffer: &Vec<u8>| {
```
ecoal95: nit: You can prevent the extra indirection using `&[UnixChannel]` instead of `&Vec<UnixChannel>`, and `&[u8]` instead of `&Vec<u8>`.

platform/linux/mod.rs
```diff
@@ -691,21 +709,19 @@ fn recv(fd: c_int, blocking_mode: BlockingMode)
             return Ok((main_data_buffer, channels, shared_memory_regions))
         }
 
-        // Reassemble fragments.
-        let mut leftover_fragments = Vec::new();
+        // Reassomble fragments.
```
ecoal95: nit: Reassemble

ecoal95:
If I understand this well, this works by creating an extra dedicated channel at send, and receiving the dedicated channel receiver at recv time, and receiving leftovers through the dedicated channel.

I don't know how performant is it, but at least it seems correct :)

FWIW, all the tests are passing locally for me in both debug and release mode.

r? @pcwalton

ecoal95:
@antrik Do you know if the max message size received in a single fragment is well defined?

Could the extra channel creation be prevented with small messages?

I guess it's not, but...

platform/linux/mod.rs
antrik: I wonder whether this would improve readability and/or performance?

(BTW, I intend to make major changes to this code in further iterations -- so probably not worthwhile to spend too much time polishing details... But I can change it of course if it's generally considered better style.)

platform/linux/mod.rs
antrik: Whoops... Not sure how this slipped through. I usually reread my diffs thoroughly to prevent this kind of fail :-(

antrik:
@ecoal95 check out the rest of the function: fragmentation (and creation of the dedicated channel) is only done if sending in one piece failed. For small messages, this change shouldn't introduce any overhead whatsoever.

platform/linux/mod.rs
ecoal95: I think it improves both (although performance-wise it's probably optimized away in release). It's fine for me if you don't change it just yet though.

ecoal95:
@antrik Whoops, sorry, that's true (didn't read thoroughly, and missed the early `return Ok(())`).

Nice then! :)

platform/linux/mod.rs
antrik: OK -- being a newbie at Rust, I'll just trust you on this :-)

platform/linux/mod.rs
antrik: I tried this too; however, it can't actually be passed uninitialised as far as I can see (the compiler complains) -- so in the end it looked uglier to me than the Box approach.

(Especially since in truth the iovec field is actually a variable-sized array... Which I intend to make explicit in one of the follow-up commits.)

antrik:
Fixed the nits -- and also removed the now unused UnixCmsg.send() method, which I forgot earlier...

platform/linux/mod.rs
ecoal95: There is `std::mem::uninitialized()` that allows you to do that :P

pcwalton:
Looks good other than one stylistic nit. Thanks so much for looking into this!

antrik:
That was the other major stylistic doubt I had... ;-)

The only variable it closes over is shared_memory_regions. I considered passing it explicitly too for more uniformity... And I also tried declaring it as a "regular" function. It felt a bit awkward because of the additional type annotations needed with that syntax -- but I'm not really opposed to it.

However, if I understand you right, you would actually want that as an entirely separate function rather than a nested one? I'd rather not do that: this code is tightly coupled; I'm sure it will feel awkward when I will be making further changes to it later on...

pcwalton:
OK, I don't really care too much either way. I think I'd prefer a nested function if it isn't too much trouble.

antrik:
Turned it into a "regular" nested function now.

antrik:
Weird: it turns out that this change actually slightly improves performance even in the non-concurrent case...

PR #42, opened 2016-02-22
Stop using the deprecated set_memory method.

Ms2ger:


jdm:
@bors-servo: r+
Not waiting for appveyor results since the changes don't modify in-process.

PR #44, opened 2016-03-04
Don't panic on Linux when shmem is zero-sized.

pcwalton:
r? @mbrubeck


platform/linux/mod.rs
```diff
@@ -32,6 +32,9 @@ const DEV_NULL_RDEV: libc::dev_t = 0x0103;
 
 const MAX_FDS_IN_CMSG: u32 = 64;
 
+// Yes, really!
+const MAP_FAILED: *mut u8 = (!0) as usize as *mut u8;
```
mbrubeck: I'd prefer to write this `!0usize as *mut u8` for clarity.

PR #47, opened 2016-03-08
Fix zero-sized shared memory leaking FD on drop (due to len=0 making …

### LINKED ISSUE #45: Intermittent assertions on Linux when using WR
I'm still seeing regular assertions on Linux when using WR - don't have a repro yet sorry. It seems to happen within a minute or so of running - perhaps a resource leak?

thread 'LayoutThread PipelineId { namespace_id: PipelineNamespaceId(0), index: PipelineIndex(0) }' panicked at 'assertion failed: fd >= 0', /home/projects/work/servo/servo/.cargo/git/checkouts/ipc-channel-3840821e54ea11a6/master/platform/linux/mod.rs:765

glennw:
…mmap fail).

Fixes #45.

### DIFF
--- a/platform/linux/mod.rs
+++ b/platform/linux/mod.rs
@@ -549,13 +549,13 @@ unsafe impl Sync for UnixSharedMemory {}
 
 impl Drop for UnixSharedMemory {
     fn drop(&mut self) {
-        if !self.ptr.is_null() {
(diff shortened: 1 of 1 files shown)


pcwalton:
Doh! Nice catch.

PR #49, opened 2016-03-09
Allow serde 0.7

nox:
We now use bincode from crates.io.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,22 +1,17 @@
 [package]
 name = "ipc-channel"
-version = "0.2.0"
+version = "0.2.1"
 
 [lib]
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -1,22 +1,17 @@
...
-version = "0.2.0"
+version = "0.2.1"
 
 [lib]
 name = "ipc_channel"
 path = "lib.rs"
 
-[dependencies.serde]
-version = "0.6"
-opt_level = 2
```
antrik: Why did you remove this?

nox: Because it doesn't do anything according to Cargo.

PR #52, opened 2016-03-18
Fix bugs in handling of socket addresses

antrik:
The main commit (the third one, about obtaining '\0'-terminated C strings) addresses a pretty serious bug, which is quite likely causing some intermittent failures. The others are minor.

This also includes a commit with general improvement to some existing test cases, which is not strictly related to the issue at hand, but helped debugging it.

### DIFF
--- a/platform/linux/mod.rs
+++ b/platform/linux/mod.rs
@@ -288,7 +288,7 @@ impl UnixSender {
             };
             libc::strncpy(sockaddr.sun_path.as_mut_ptr(),
                           name.as_ptr(),
-                          sockaddr.sun_path.len() as size_t);
+                          sockaddr.sun_path.len() as size_t - 1);
 
             let len = mem::size_of::<c_short>() +
                 (libc::strlen(sockaddr.sun_path.as_ptr()) as usize);
@@ -469,7 +469,7 @@ impl UnixOneShotServer {
                 };
                 libc::strncpy(sockaddr.sun_path.as_mut_ptr(),
                               path.as_ptr() as *const c_char,
-                              sockaddr.sun_path.len() as size_t);
+                              sockaddr.sun_path.len() as size_t - 1);
 
                 let len = mem::size_of::<c_short>() + (libc::strlen(sockaddr.sun_path.as_ptr()) as
                                                        usize);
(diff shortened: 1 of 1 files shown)


test.rs
```diff
@@ -154,20 +156,25 @@ fn cross_process_embedded_senders() {
...
+            0 => {
+                let (tx1, rx1): (IpcSender<Person>, IpcReceiver<Person>) = ipc::channel().unwrap();
+                let tx0 = IpcSender::connect(server0_name).unwrap();
+                tx0.send(tx1).unwrap();
+                rx1.recv().unwrap();
+                let tx2: IpcSender<Person> = IpcSender::connect(server2_name).unwrap();
+                tx2.send(person.clone()).unwrap();
+                libc::exit(0);
+            },
+            pid => pid,
```
pcwalton: Maybe wrap these up in a helper function?

platform/linux/mod.rs
```diff
@@ -459,7 +459,7 @@ impl UnixOneShotServer {
             loop {
                 let path_string = CString::new(b"/tmp/rust-ipc-socket.XXXXXX" as &[u8]).unwrap();
                 path = path_string.as_bytes().iter().cloned().collect();
-                if mktemp(path.as_mut_ptr() as *mut c_char) == ptr::null_mut() {
+                if *mktemp(path.as_mut_ptr() as *mut c_char) == 0 {
```
pcwalton: Ugh! Nice catch.

test.rs
antrik: Hm... Makes sense I guess. Where shall I put it though? It's used both in tests.rs and in platform/tests.rs...

pcwalton: `tests.rs` would be fine.

antrik:
I dropped the tests change from this PR, so it doesn't hold up the important fix. I'll submit it separately.

test.rs
antrik: Eh, actually I meant `test.rs` and `platform/test.rs`. So when you say `tests.rs`, do you actually mean `test.rs` as well -- or do you mean I should create a separate file for that?...

PR #53, opened 2016-03-19
tests: Wait for forked child processes

antrik:
Don't leave zombie processes around -- they hinder debugging, and might
have other harmful side effects as well.

### DIFF
--- a/test.rs
+++ b/test.rs
@@ -11,10 +11,37 @@ use ipc::{self, IpcOneShotServer, IpcReceiver, IpcReceiverSet, IpcSender, IpcSha
 use ipc::{OpaqueIpcSender};
 use router::ROUTER;
 use libc;
+use std::io::Error;
 use std::iter;
+use std::ptr;
 use std::sync::Arc;
 use std::sync::mpsc::{self, Sender};
 use std::thread;
+
+///XXXjdm Windows' libc doesn't include fork.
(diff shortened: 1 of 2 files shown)


antrik:
Refactored variant. Let's hope I didn't break Windows build...

pcwalton:
I don't think it's unsafe, but it doesn't really matter.

test.rs
```diff
@@ -11,11 +11,38 @@ use ipc::{self, IpcOneShotServer, IpcReceiver, IpcReceiverSet, IpcSender, IpcSha
...
 use std::iter;
+use std::ptr;
 use std::sync::Arc;
 use std::sync::mpsc::{self, Sender};
 use std::thread;
 
+///XXXjdm Windows' libc doesn't include fork.
+#[cfg(not(windows))]
+// I'm not actually sure invoking this is indeed unsafe -- but better safe than sorry...
+pub unsafe fn fork<F: FnOnce()>(child_func: F) -> libc::pid_t {
```
nox: I think `F` should be `FnOnce() -> !`, to show that it shouldn't return ever.

antrik: For the record: I actually think so too; but because of language limitations, this can currently be only achieved by using the Void crate -- which @pcwalton didn't want.

PR #56, opened 2016-03-24
Fix two leaks in the Mac platform code.

pcwalton:
Related: servo/servo#10029

r? @metajack


platform/macos/mod.rs
```diff
@@ -591,6 +591,8 @@ fn select(port: mach_port_t, blocking_mode: BlockingMode)
                         os_result => return Err(MachError(os_result)),
```
metajack: Can we leak here in the case that we get a bad os_result?

metajack:
Looks good aside from the one comment.

pcwalton:
Fixed. r? @metajack

PR #57, opened 2016-03-28
Limit Linux sockets to 32kB buffers.

eddyb:
Based on [the code in `gonk/libui`](https://dxr.mozilla.org/mozilla-central/rev/63be002b4a803df1122823841ef7633b7561d873/widget/gonk/libui/InputTransport.cpp#130-134), this appears to fix servo/servo#9555.

### DIFF
--- a/platform/linux/mod.rs
+++ b/platform/linux/mod.rs
@@ -34,7 +34,20 @@ pub fn channel() -> Result<(UnixSender, UnixReceiver),UnixError> {
     let mut results = [0, 0];
     unsafe {
         if socketpair(libc::AF_UNIX, SOCK_SEQPACKET, 0, &mut results[0]) >= 0 {
-            Ok((UnixSender::from_fd(results[0]), UnixReceiver::from_fd(results[1])))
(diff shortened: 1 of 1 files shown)


platform/linux/mod.rs
```diff
@@ -34,7 +34,20 @@ pub fn channel() -> Result<(UnixSender, UnixReceiver),UnixError> {
     let mut results = [0, 0];
     unsafe {
         if socketpair(libc::AF_UNIX, SOCK_SEQPACKET, 0, &mut results[0]) >= 0 {
-            Ok((UnixSender::from_fd(results[0]), UnixReceiver::from_fd(results[1])))
+            let (a, b) = (results[0], results[1]);
+
+            // Limit the buffers to 32kB.
+            let buf_sz: c_int = 32 * 1024;
```
pcwalton: Can you factor this out into a top-level constant?

pcwalton:
@bors-servo: r+

It's really annoying that this is necessary, but…

Issue #33, opened 2016-01-30
Write a test for #30


jdm:
Should be straightforward - make a channel, send a value on the sender, drop the sender, and try to receive it from the receiver.

PR #30, opened 2016-01-22
Turn on `SO_LINGER` for client communication sockets.

pcwalton:
This reduces the probability that the receiver receives errors when we
close our end of the socket with data remaining.

There is deadlock potential with this patch, because turning on `SO_LINGER`
causes `close()` to block until the receiver has received all the data. If
deadlocks happen, a workaround will be to close sockets in a separate thread.
This is ugly and slow, so I don't want to do that unless we need to.

Might fix #29.

r? @jdm

PR #58, opened 2016-03-28
Test: data is recv-able after sender is dropped

danlrobertson:
Add test for #30. Test that data can be retrieved from the receiver
after the sender has been dropped.

Fixes: #33

---
This change is [Reviewable](https://reviewable.io/reviews/servo/ipc-channel/58)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 *~
+*.swp
+*.swo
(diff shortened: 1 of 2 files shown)


jdm:
@bors-servo: r+
Thanks!

PR #59, opened 2016-04-01
Don't set SO_RCVBUF on UNIX sockets.

eddyb:
Fixes servo/servo#10260, although this might not be even a reasonable approach.
At this point it's mostly guesswork and Servo needs a lot more testing on linux.


pcwalton:
Bluh?

I mean, r=me I guess, but this is kinda scary…

samlh:
Pure speculation: if there are multiple senders, then perhaps the receive buffer needs to be large enough to receive from all simultaneously?

eddyb:
Closed in favor of #60.

PR #60, opened 2016-04-01
Linux: UnixCmsg: Don't deduce maximum cmsg size from data size

eddyb:
@antrik committed 10f4032:
> Just because we allocate a generous receive buffer for auxilllary data,
> doesn't mean that a message can't use the space for other things when it
> doesn't actually carry that amount of auxillary data.
> 
> In fact, there is no reason to expect every message we receive to have
> auxillary data at all -- so let's just allocate a data buffer matching
> the total maximal receive size. This will enable some optimisations to
> be implemented; and it simplifies the code too.
> 
> Note that the extra space is not used yet, until the sender side is
> adapted to actually send larger messages making use of the extra space.
> (Or even messages without auxillary data.)

This also fixes servo/servo#10260, but unlike #59, it's not just a hack.
I've put this up as a PR because @antrik might not get to it for a few days.
cc @Manishearth @pcwalton

### DIFF
--- a/platform/linux/mod.rs
+++ b/platform/linux/mod.rs
@@ -806,12 +806,11 @@ impl UnixCmsg {
         let cmsg_length = mem::size_of::<cmsghdr>() + (MAX_FDS_IN_CMSG as usize) *
             mem::size_of::<c_int>();
         assert!(maximum_recv_size > cmsg_length);
-        let data_length = maximum_recv_size - cmsg_length;
-        let mut data_buffer: Vec<u8> = vec![0; data_length];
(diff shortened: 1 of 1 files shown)


Manishearth:
This works for me; I'm not getting any IPC related crashes.

antrik:
If this can wait till later today or maybe tomorrow, I'd put up a more polished PR: with a reworked commit message mentioning the bug fix; and adding a test case... It's not too important though, if this fix is urgent.

antrik:
Please close in favour of https://github.com/servo/ipc-channel/pull/61

PR #62, opened 2016-04-03
Handle ENOBUFS by dynamically reducing packet size

antrik:
This is a better fix for https://github.com/servo/servo/issues/9555 : handling the problem when it actually occurs, without affecting performance otherwise.

It obsoletes (and removes) the workaround in https://github.com/servo/ipc-channel/pull/57

### DIFF
--- a/platform/linux/mod.rs
+++ b/platform/linux/mod.rs
@@ -25,9 +25,6 @@ use std::thread;
 
 const MAX_FDS_IN_CMSG: u32 = 64;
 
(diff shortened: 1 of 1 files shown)


eddyb:
Works for me, both with and without `browser.html`. cc @Manishearth @pcwalton

Manishearth:
Same

pcwalton:
@bors-servo: r+

Holy cow, Linux can be terrible sometimes.

PR #68, opened 2016-05-01
Optimisations (and cleanups) of Linux backend

antrik:
This is a bunch of optimisations to the Linux platform code, along with various cleanups of the related code which the optimisation patches build upon.

Most of the cleanups are on the `send()` side, as the `recv()` side is less affected by the optimisation changes, and thus there has been less reason to refactor the code -- some extra cleanup work would probably be in order here.

The optimisations are mostly about avoiding unnecessary copies by using scatter-gather buffers for send and receive; as well as avoiding unnecessary initialisation of receive buffers.

The results are impressive: with gains of at least 5x for large transfers of several MiB (a bit more on a modern system); >5x (on an old system) up to >10x (on a modern one) for small transfers of up to a few KiB; and more than 10x for most of the range in between -- peaking at about 12x - 13x on the old system and 20x - 21x on the modern system for medium-sized transfers of about 64 KiB up to a few hundred KiB.

For another interesting data point, the CPU usage during benchmark runs (with many iterations, to amortise the setup time) was dominated by user time (more than two thirds of total time) with the original variant; whereas the optimised variant not only further reduces system time to less then half the original value (presumably because of fewer allocations?), but also almost entirely eliminates the user time, making it pretty insignificant in the total picture now -- as it should be.

On a less scientific note, Servo built with the optimised ipc-channel doesn't seem to show undue delays any more while rendering a language selection menu. (Which requires lots of fonts to be loaded, and thus triggers heavy ipc-channel activity.)

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -25,7 +25,7 @@ fn bench_size(b: &mut test::Bencher, size: usize) {
     let (wait_tx, wait_rx) = mpsc::channel();
     let wait_rx = Mutex::new(wait_rx);
 
-    if size > tx.get_max_fragment_size().unwrap() {
+    if size > platform::OsIpcSender::get_max_fragment_size() {
         b.iter(|| {
             crossbeam::scope(|scope| {
                 scope.spawn(|| {
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -1,0 +1,183 @@
+#![feature(test)]
+
+extern crate crossbeam;
+extern crate ipc_channel;
+extern crate test;
+
+use ipc_channel::platform;
+
+use std::sync::{mpsc, Mutex};
+
+/// Allows doing multiple inner iterations per bench.iter() run.
+///
+/// This is mostly to amortise the overhead of spawning a thread in the benchmark
+/// when sending larger messages (that might be fragmented).
+///
+/// Note that you need to compensate the displayed results
+/// for the proportionally longer runs yourself,
+/// as the benchmark framework doesn't know about the inner iterations...
+const ITERATIONS: usize = 1;
+
+fn bench_size(b: &mut test::Bencher, size: usize) {
+    let data: Vec<u8> = (0..size).map(|i| (i % 251) as u8).collect();
+    let (tx, rx) = platform::channel().unwrap();
+
+    let (wait_tx, wait_rx) = mpsc::channel();
+    let wait_rx = Mutex::new(wait_rx);
+
+    if size > tx.get_max_fragment_size().unwrap() {
+        b.iter(|| {
+            crossbeam::scope(|scope| {
+                scope.spawn(|| {
+                    let wait_rx = wait_rx.lock().unwrap();
+                    for _ in 0..ITERATIONS {
+                        tx.send(&data, vec![], vec![]).unwrap();
+                        if ITERATIONS > 1 {
+                            // Prevent beginning of the next send
+                            // from overlapping with receive of last fragment,
+                            // as otherwise results of runs with a large tail fragment
+                            // are significantly skewed.
+                            wait_rx.recv().unwrap();
+                        }
+                    }
+                });
+                for _ in 0..ITERATIONS {
+                    rx.recv().unwrap();
+                    if ITERATIONS > 1 {
+                        wait_tx.send(()).unwrap();
+                    }
+                }
+                // For reasons mysterious to me,
+                // not returning a value *from every branch*
+                // adds some 100 ns or so of overhead to all results --
+                // which is quite significant for very short tests...
+                0
+            })
+        });
+    } else {
+        b.iter(|| {
+            for _ in 0..ITERATIONS {
+                tx.send(&data, vec![], vec![]).unwrap();
+                rx.recv().unwrap();
+            }
+            0
+        });
+    }
+}
+
+// It turns out the results have some crazy jumps between sizes,
+// probably related to some allocator alignment stuff.
+// What's more, these fluctuate strongly in seemingly random ways
+// in response to various changes to the test setup etc.
+//
+// Warming up the memory somewhat mitigates these issues.
+// The specific size used here was determined empirically
+// to produce comparatively sane results -- on my system at least.
+// (Maybe because it doesn't align well with 2's complements...
+// Or maybe it's just an entirely random effect.)
+//
+// There might be more elegant and/or more effective ways to do the warm-up,
+// using random allocations or something along these lines...
+// However, I don't think it's really *that* important --
+// and I already spent way too much time trying to figure this out :-(
+#[bench]
+fn _invoke_chaos(b: &mut test::Bencher) {
+    bench_size(b, 777_777); // 111-up the beast ;-)
+}
+
+#[bench]
+fn size_00_1(b: &mut test::Bencher) {
+    bench_size(b, 1);
+}
+#[bench]
+fn size_01_2(b: &mut test::Bencher) {
+    bench_size(b, 2);
+}
+#[bench]
+fn size_02_4(b: &mut test::Bencher) {
+    bench_size(b, 4);
+}
+#[bench]
+fn size_03_8(b: &mut test::Bencher) {
+    bench_size(b, 8);
+}
+#[bench]
+fn size_04_16(b: &mut test::Bencher) {
+    bench_size(b, 16);
+}
+#[bench]
+fn size_05_32(b: &mut test::Bencher) {
+    bench_size(b, 32);
+}
+#[bench]
+fn size_06_64(b: &mut test::Bencher) {
+    bench_size(b, 64);
+}
+#[bench]
+fn size_07_128(b: &mut test::Bencher) {
+    bench_size(b, 128);
+}
+#[bench]
+fn size_08_256(b: &mut test::Bencher) {
+    bench_size(b, 256);
+}
+#[bench]
+fn size_09_512(b: &mut test::Bencher) {
+    bench_size(b, 512);
+}
+#[bench]
+fn size_10_1k(b: &mut test::Bencher) {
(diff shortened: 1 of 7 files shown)


jdm:
Build failure is from aster and quasi crates breaking on the current rustc nightly.

antrik:
I now submitted a bunch of PRs for this: https://github.com/serde-rs/syntex/pull/44 , https://github.com/serde-rs/aster/pull/78 , https://github.com/serde-rs/quasi/pull/41 , https://github.com/serde-rs/serde/pull/303

Once these are in, we can update the dependencies in ipc-channel...

platform/linux/mod.rs
```diff
@@ -278,16 +286,10 @@ impl UnixSender {
 
                 if result <= 0 {
                     let error = UnixError::last();
-                    if error.0 == libc::ENOBUFS && bytes_to_send > 2000 {
+                    if error.0 == libc::ENOBUFS
+                       && downsize(&mut sendbuf_size, bytes_to_send).is_ok() {
```
samlh: This does not handle the Err(()) case from downsize correctly - perhaps this should be:
```rust
if error.0 == libc::ENOBUFS {
    // If the kernel failed to allocate a buffer large enough for the packet,
    // retry with a smaller size (if possible).
    try!(downsize(&mut sendbuf_size, bytes_to_send));
    continue;
} else {
```

Note this may need a tweak for freeing behavior.

There is a similar issue above, but a later commit makes it moot.

platform/linux/mod.rs
```diff
@@ -200,27 +201,33 @@ impl UnixSender {
...
                     iov_base: data_buffer.as_ptr() as *const c_char as *mut c_char,
                     iov_len: data_buffer.len(),
-                });
-
+                };
                 let msghdr = msghdr {
                     msg_name: ptr::null_mut(),
                     msg_namelen: 0,
-                    msg_iov: &mut *iovec,
+                    msg_iov: &iovec as *const _ as *mut _,
```
Ms2ger: This seems fishy. Why not keep it `mut`?

platform/linux/mod.rs
antrik: I don't think returning the empty error from `downsize()` would be correct handling... The idea here is that if the downsize fails, we want to return the original error -- which happens in the else case.

As I said in the commit message, I'm not entirely happy how this turned out syntactically -- but I really can't think of a better approach. Quite frankly, I considered dropping this size test entirely: it's a case that I don't actually expect to happen; I only added it as an additional safeguard in the original `ENOBUFS` handling because it seemed easy to do -- but it turned out to be a major pain during refactoring, and I'm not sure any more it's indeed worth the trouble...

Maybe we should just turn it into an `assert()` instead?

platform/linux/mod.rs
antrik: The idea was that `sendmsg()` only reads data, rather than mutating, and thus it makes sense to also make the `iovec` immutable here.

Later I realised that the `iovec` itself (as opposed to the buffers it references) is never mutated in the `recvmsg()` either -- and indeed it's declared as `const` in the definition of the same structure in libc. So I considered adjusting it on the receiver side too -- but ultimately decided not to bother, as one of the things I intend to do next is switching to the libc definitions anyway.

Maybe I should add this cleanup here after all, to avoid confusion...

antrik:
Added a patch making `msghdr.iovec` a `*const` too avoid confusion. (And did a few other minor cleanups to `iovec` handling along the way...)

platform/linux/mod.rs
samlh: Fair enough. I was looking at the call above that has an else if instead of an else, but that is fixed by a later commit. Thanks for taking a look, either solution works for me.

platform/test.rs
```diff
@@ -174,30 +173,77 @@ fn big_data_with_sender_transfer() {
     thread.join().unwrap();
 }
 
-#[test]
-// This test only applies to platforms that need fragmentation.
+macro_rules! create_with_n_fds {
+    ($name:ident, $n:expr, $size:expr) => (
+        #[test]
+        fn $name() {
```
pcwalton: Does this really need to be a macro? Can it just be a helper function that the various tests delegate to?

"Macros are for when you have run out of language" —@littlecalculist

But I think there is still language left :)

platform/linux/mod.rs
```diff
@@ -227,8 +227,7 @@ impl UnixSender {
             let (msghdr, mut iovec) = construct_header(&fds[..], &data_buffer[..]);
```
pcwalton: typo in commit message: "auxiliary"

platform/linux/mod.rs
```diff
@@ -28,6 +28,10 @@ const MAX_FDS_IN_CMSG: u32 = 64;
 // Yes, really!
 const MAP_FAILED: *mut u8 = (!0usize) as *mut u8;
 
+// The value Linux returns for SO_SNDBUF
+// is not the size we are actually allowed to use...
+const RESERVED_SIZE: usize = 256;
```
pcwalton: Can you elaborate in the comment as to how this is derived? This is kinda scary.

platform/linux/mod.rs
```diff
@@ -30,7 +30,8 @@ const MAP_FAILED: *mut u8 = (!0usize) as *mut u8;
 
 // The value Linux returns for SO_SNDBUF
 // is not the size we are actually allowed to use...
-const RESERVED_SIZE: usize = 256;
+// Empirically, we have to deduct 32 bytes from that.
+const RESERVED_SIZE: usize = 32;
```
pcwalton: Super weird.

platform/linux/mod.rs
```diff
@@ -138,6 +138,21 @@ impl UnixSender {
         }
     }
 
+    /// Calculate maximum payload data size per fragment.
+    ///
+    /// This is the size of the main data chunk only --
+    /// it's independent of any auxillary data (FDs) transferred along with it.
```
pcwalton: typo: auxiliary

platform/linux/mod.rs
```diff
@@ -153,6 +153,19 @@ impl UnixSender {
         sendbuf_size - RESERVED_SIZE - mem::size_of::<u32>() * 2
     }
 
+    /// Maximum data size that can be transferred over this channel in a single packet.
+    ///
+    /// This is the size of the main data chunk only --
+    /// it's independent of any auxillary data (FDs) transferred along with it.
```
pcwalton: typo: auxiliary

benches/bench.rs
```diff
@@ -0,0 +1,113 @@
...
+///
+/// Note that you need to compensate the displayed results
+/// for the proportionally longer runs yourself,
+/// as the benchmark framework doesn't know about the inner iterations...
+const ITERATIONS: usize = 1;
+
+macro_rules! create_bench_for_size {
+    ($name:ident, $size:expr) => (
+        #[bench]
+        fn $name(b: &mut test::Bencher) {
```
pcwalton: Can this just be a helper function instead of a macro?

platform/linux/mod.rs
```diff
@@ -290,30 +291,24 @@ impl UnixSender {
...
-                data_buffer.write_u32::<LittleEndian>(next_fragment_id).unwrap();
-                data_buffer.write(&data[byte_position..end_byte_position]).unwrap();
-            }
-
-            let bytes_to_send = end_byte_position - byte_position + mem::size_of::<u32>() * 2;
 
+            let bytes_to_send;
             let result = if byte_position == 0 {
+                // First fragment. No offset; but contains message header (total size).
+                // The auxillary data (FDs) is also sent along with this one.
```
pcwalton: typo: auxiliary

platform/test.rs
antrik: Well, I did use a helper function at first; but that means there were a lot of redundant wrappers for each individual test. I didn't like it.

(The macro instantiations admittedly still involve some redundancy for now, due to limitations of the current macro system...)

BTW, I introduced a similar macro already in an earlier PR -- you didn't comment on that one...

pcwalton: Redundant wrappers? What do you mean?

I don't see how wrapper functions are redundant. There's a tiny amount of boilerplate, but the clarity is worth it IMO. I always have a hard time reading macros.

platform/linux/mod.rs
antrik: Well, I took it from your original code -- so it's up to you to explain it :-P

I thought that it might have been related to `MAX_FDS_IN_CMSG`, which would match the size -- but then, you were also deducting space depending on the number of FDs *actually* sent; so that doesn't really make sense either...

Anyway, as you have seen, I changed in the next commit.

metajack: > Well, I took it from your original code -- so it's up to you to explain it :-P

That made me laugh. Maybe self-review isn't so bad after all.

platform/linux/mod.rs
antrik: Indeed.

The man page actually mentions the buffer being *twice* the usable size, but in reference to a different socket type. There is no documentation about this for `SOCK_SEQPACKET` as far as I'm aware.

platform/linux/mod.rs
```diff
@@ -194,14 +194,19 @@ impl UnixSender {
                                -> Result<(),UnixError> {
```
pcwalton: typo in commit message: "auxiliary"

pcwalton:
This looks good to me with the above nits addressed. Thanks!

antrik:
Rebased; fixed typos; replaced macros by helper functions + boilerplate. That should cover everything I hope?...

(Let's see whether CI succeeds on non-Linux, now that the Serde issues are sorted out...)

antrik:
OK, let's try again...

PR #77, opened 2016-05-23
Put file descriptors in an Arc

Manishearth:
We used to do this manually in servo, but @izgzhen turned it off when consolidating the resource senders

There really isn't much of a reason to not do this; and `mpsc::Sender` does this already (for a different reason)

r? @antrik or @pcwalton

cc @asajeffrey

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ipc-channel"
-version = "0.2.2"
+version = "0.2.3"
 
 [lib]
 name = "ipc_channel"
--- a/platform/linux/mod.rs
+++ b/platform/linux/mod.rs
@@ -19,6 +19,7 @@ use std::mem;
 use std::ops::Deref;
 use std::ptr;
 use std::slice;
+use std::sync::Arc;
 use std::thread;
 
 const MAX_FDS_IN_CMSG: u32 = 64;
@@ -91,26 +92,20 @@ impl UnixReceiver {
     }
 }
 
-#[derive(PartialEq, Debug)]
+#[derive(PartialEq, Debug, Clone)]
+pub struct SharedFileDescriptor(c_int);
+
+
+#[derive(PartialEq, Debug, Clone)]
 pub struct UnixSender {
-    fd: c_int,
+    fd: Arc<SharedFileDescriptor>,
 }
 
-impl Drop for UnixSender {
+impl Drop for SharedFileDescriptor {
     fn drop(&mut self) {
         unsafe {
-            let result = libc::close(self.fd);
+            let result = libc::close(self.0);
             assert!(thread::panicking() || result == 0);
-        }
-    }
-}
-
-impl Clone for UnixSender {
-    fn clone(&self) -> UnixSender {
-        unsafe {
-            UnixSender {
-                fd: libc::dup(self.fd)
-            }
         }
     }
 }
@@ -118,7 +113,7 @@ impl Clone for UnixSender {
 impl UnixSender {
     fn from_fd(fd: c_int) -> UnixSender {
         UnixSender {
-            fd: fd,
+            fd: Arc::new(SharedFileDescriptor(fd)),
         }
     }
 
@@ -130,7 +125,7 @@ impl UnixSender {
         unsafe {
             let mut socket_sendbuf_size: usize = 0;
             let mut socket_sendbuf_size_len = mem::size_of::<usize>() as socklen_t;
-            if getsockopt(self.fd,
+            if getsockopt(self.fd.0,
                           libc::SOL_SOCKET,
                           libc::SO_SNDBUF,
                           &mut socket_sendbuf_size as *mut _ as *mut c_void,
@@ -289,7 +284,7 @@ impl UnixSender {
 
         // If the message is small enough, try sending it in a single fragment.
         if data.len() <= Self::get_max_fragment_size() {
-            match send_first_fragment(self.fd, &fds[..], data, data.len()) {
+            match send_first_fragment(self.fd.0, &fds[..], data, data.len()) {
                 Ok(_) => return Ok(()),
                 Err(error) => {
                     // ENOBUFS means the kernel failed to allocate a buffer large enough
@@ -328,13 +323,13 @@ impl UnixSender {
 
                 // This fragment always uses the full allowable buffer size.
                 end_byte_position = Self::first_fragment_size(sendbuf_size);
-                send_first_fragment(self.fd, &fds[..], &data[..end_byte_position], data.len())
+                send_first_fragment(self.fd.0, &fds[..], &data[..end_byte_position], data.len())
             } else {
                 // Followup fragment. No header; but offset by amount of data already sent.
 
                 end_byte_position = cmp::min(byte_position + Self::fragment_size(sendbuf_size),
                                              data.len());
-                send_followup_fragment(dedicated_tx.fd, &data[byte_position..end_byte_position])
+                send_followup_fragment(dedicated_tx.fd.0, &data[byte_position..end_byte_position])
             };
 
             if let Err(error) = result {
@@ -385,7 +380,7 @@ pub enum UnixChannel {
 impl UnixChannel {
     fn fd(&self) -> c_int {
         match *self {
-            UnixChannel::Sender(ref sender) => sender.fd,
+            UnixChannel::Sender(ref sender) => sender.fd.0,
             UnixChannel::Receiver(ref receiver) => receiver.fd,
         }
     }


asajeffrey:
IRC chat: http://logs.glob.uno/?c=mozilla%23servo&s=23+May+2016&e=23+May+2016#c436462

Summary: I'm not sure how much this helps in the multi-process case, or even in the single-process case where `IPCSender`s are serialized/deserialized a lot.

Manishearth:
We had `Arc<ResourceThread>` up until recently which was there to prevent running out of fds. Zhen removed it and we started getting the "too many open files" errors again. It fixes a real error right now that breaks CI.

metajack:
Why did @izgzhen turn it off? Or was it just an unintentional byproduct of his patch?

izgzhen:
@metajack It is mostly like a intentional byproduct. Not all members of `ResourceThreads` are `Arc`'d, and the change (from plain-old `IpcSender` to an Arc'd version) only happens on the half-way between initialisation and actual use, which makes the interface very hard to prose and pass the compiler check.

izgzhen:
@metajack It is mostly like a intentional byproduct. Not all members of `ResourceThreads` are `Arc`'d, and the change (from plain-old `IpcSender` to an Arc'd version) only happens on the half-way between initialisation and actual use, which makes the interface very hard to prose and pass the compiler check.

antrik:
I'm ambivalent about this: on one hand, intuitively it seems wrong to add an extra layer of indirection for everyone, just because of a single user running into problems...

(For the specific issue @izgzhen faced in https://github.com/servo/servo/pull/11189 , it seems to me the solution would be for `script-thread` simply not to use the "consolidated" `ResourceThreads` struct;  and instead just to keep a separate `Arc<CoreResourceThread>` variable, as it needs to handle it explicitly anyway when passing it to `DocumentLoader`. Quite frankly, I'm not very fond of how `resource_threads.sender()` returns something different depending on an inferred type parameter anyway... It's very opaque IMHO.)

On the other hand, I see that the various `Thread` structs get cloned in various other places too; and while they might not be as critical, they cause unnecessary resource usage as well. Of course they could all be changed to use `Arc`s; but then, since it seems a common pattern -- and there is really no good reason ever to clone the actual `OsIpcSender` structure as far as I'm aware -- it can very well be argued that it's indeed better just to put the `Arc` into the `Sender` itself.

One thing to consider though is that for the `inprocess` platform backend, the `OsIpcSender` is actually an `mpsc::Sender`; so wrapping it in another `Arc` is pointless I guess? Thus probably it would be better to put the `Arc` in the `UnixSender` (in the Linux platform backend) instead. (And probably the Mach (MacOS) backend too: while cloning Mach ports should be much cheaper that cloning FDs, I guess it's still unnecessary overhead...)

metajack:
I'd like @pcwalton's input on this.

@Manishearth what do you think about the alternative that @antrik mentioned?

If mpsc::Sender is an Arc under the hood, it seems like making IPCSender an Arc (except not doubling it for in-process) would be fairly consistent with that design. But I am not as familiar with the context here as others on the team.

antrik:
Thinking about it, `UnixSender` is used in some places "internally" that definitely don't need the `Arc`... So actually I'm not sure about the best place to put it.

ipc.rs
```diff
@@ -179,7 +180,7 @@ impl<T> IpcSender<T> where T: Serialize {
 
     pub fn to_opaque(self) -> OpaqueIpcSender {
         OpaqueIpcSender {
-            os_sender: self.os_sender,
+            os_sender: self.os_sender.clone(),
```
antrik: Why do you need to clone here? Isn't self consumed anyway? Or am I missing something?...

Manishearth: My bad, fixed

Manishearth:
@antrik does this look better to you? I'll look at the osx one later

antrik:
@Manishearth this is kinda funny: most of this change is actually about creating a dedicated FD type, and moving the `Clone` and `Drop` implementations there -- which is *precisely* what I was about to do this very moment, though for other reasons :-)

Indeed my *first* thought (when I was originally thinking about it a while back) was to put it in an `Arc` -- only later I concluded that it shouldn't *strictly* be necessary...

(Note: the title of the PR doesn't really fit any more.)

Manishearth:
heh.

So should I move forward with this and Arcify the mac fds as well?

antrik:
Nitpick: I wonder whether it's appropriate to call it `SharedFileDescriptor`? It seems to me there is nothing inherently shared about it -- you only make it shared in `UnixSender` by wrapping an `Arc` around it... My plan was to go with just `FileDescriptor`. Might be a matter of what angle you look at it from, though...

(Also, my intention was to use the explicit `FileDescriptor` type in `UnixReceiver` etc. as well -- even though this one has no `Clone` -- for consistency, and because it seems more elegant in general... But this requires some extra consideration; and it''s not really related to your change -- so I guess I can just do a follow-up PR later on.)

Manishearth:
FileDescriptor works

Manishearth:
Actually, we don't need this change in macos, since mach seems to have it's own refcounting api. So this is ready to merge?

pcwalton:
I considered doing this, but I decided against it because of fears of too many allocations. I'm open to it if it actually fixes stuff, though. Do we know for sure that this fixes problems?

Manishearth:
We were already doing this for the resource channel in Servo; I'd like to do it uniformly everywhere. I think the small allocation of an int is shadowed by the IPC buffer and open file costs.

Manishearth:
Right now on my servo checkout some sites do hit the file descriptor limit (which is ~1000 currently for me) if you navigate for a bit

antrik:
@pcwalton well, it's rather clear that cloning of FDs needs to be avoided -- both because of the somewhat relevant overhead, and because of the fixed number of FDs per process on many systems causing serious problems.

What's the best way to achieve that? It's a bit of a tradeoff: doing it at here will add some (very little) unnecessary overhead in cases that definitely won't clone the `Sender`. The alternative however is to add the `Arc`s explicitly for all users that might ever clone them, which is way more complicated and error-prone; and on platforms that don't actually need the wrappers, the explicitly added ones would be redundant, thus just shifting the overhead to other use cases... So the defensive programmer is clearly in favour of doing it here; while even the perfectionist and optimisation freak should be swayed by the last point.

Manishearth:
We used to have the perfectionist approach, but we only did it in a couple of places and the too many fds error still popped up.

And again, I feel like we're worrying about a really tiny allocation in face of the cost of an IPC buffer.

antrik:
To my great surprise, this does actually add a (barely) measurable overhead to channel creation (+destruction): adding about 1% to the 4.5 µs these operations take on my system. That's not much in the grand scheme of things (it's some 100 processor clock cycles); yet this is in the same ballpark as a gratuitous FD clone + drop... I for one didn't expect that.

(This is only creation, though -- I haven't tested how long actual cloning takes...)

Note that this is an oldish system (from 2008 or so) running 32 bit Linux -- I wonder whether atomic operations are more efficient on newer systems? (Unfortunately, I don't have access to a more modern system for comparison anymore :-( )

Manishearth:
Hm, interesting. Could you share your benchmarking code so I can try running it?

Note that the primary goal here is to avoid running out of fds, not avoid the overhead of an fd clone. So this being as expensive as an fd clone isn't a big deal.

antrik:
@Manishearth yeah, was just doing that right now: https://github.com/servo/ipc-channel/pull/78 :-)

I don't think this result significantly changes the situation; I mostly mentioned it because it surprised me...

Manishearth:
Mine goes from 11.4 microseconds to 11.7 microseconds. This is a pretty fast machine, not sure why this is happening. But the additional cost isn't much

Manishearth:
So, should we merge this?

antrik:
@Manishearth note that on my system at least, I need to increase ITERATIONS in bench.rs to 10000 to make the difference visible among the noise -- and even then, it requires looking at several runs of each variant to get a meaningful picture...

Regardless of that, it seems really strange that it's so much slower on you system.

Manishearth:
So I got it to happen at 2 microseconds as well but I don't know how. 

I have a bunch of things going on in the background though.

antrik:
BTW, just in case it isn't clear: I'm in favour of merging this. (Though I can't offer a full review, as I'm not familiar with the platform-agnostic serialisation layer, and how it might be affected by such a change...)

Manishearth:
Ah, okay.

@pcwalton r? on the serialization bit?

Arcs just get recreated across processes after serialization with the refcount reset, so that's okay.

Manishearth:
Yeah, with all my builds stopped and my browser closed I get 2.74/2.81 microseconds.

antrik:
@Manishearth yeah, background activity can certainly skew the results -- especially if it involves kernel load. (Creating the IPC channel is almost exclusively using kernel time, for creating the actual socket.)

Also relevant could be CPU frequency scaling. In my experience it doesn't affect results too much, but in theory it might in corner cases...

antrik:
@Manishearth these latter results certainly sound more like what I'd expect from a modern system :-) I'm a bit shocked though that creating the `Arc` appears to be even more expensive than on my old one, though...

Manishearth:
Bumping up the number of iterations, it seems like it's 4% or so.

pcwalton:
@bors-servo: r+

I'm convinced.

Manishearth:
Huh, why isn't it picking up aster 17?

Manishearth:
(working on getting the aster/quasi/syntex stuff updated)

Manishearth:
The build seemed to pass, but it didn't do anything

Manishearth:
@bors retry clean

oy vey, Appveyor wasn't fully set up on this

PR #79, opened 2016-06-01
platform/tests: cleanup: drop `truncate()` calls on received data

antrik:
I really can't think of any purpose these might serve...

### DIFF
--- a/platform/macos/mod.rs
+++ b/platform/macos/mod.rs
@@ -808,17 +808,10 @@ struct Message {
 
 impl Message {
     fn size_of(data_length: usize, port_length: usize, shared_memory_length: usize) -> usize {
-        let mut size = mem::size_of::<Message>() +
+        mem::size_of::<Message>() +
             mem::size_of::<mach_msg_port_descriptor_t>() * port_length +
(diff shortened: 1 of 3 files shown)


emilio:
Looking at the [output of the build](https://travis-ci.org/servo/ipc-channel/jobs/134534644), it seems `truncate()` makes a difference on OSX.

antrik:
Interesting... Let me check something.

antrik:
Yeah, guessed so... So normally it doesn't matter that the `macos` backend introduces trailing zeros, as they are simply ignored during deserialisation; but with `bytes_channel`, this is now exposed to the public...

Now let's try another thing...

antrik:
Meh. No idea how to fix this.

emilio:
@antrik: I'm closing this since it seems inactive. Feel free to reopen if you find a solution :/

antrik:
Heh, funny timing: I wanted to close it myself right now, as not being an Apple worshipper err I mean developer, there is nothing I can do about this...

I opened a bug report instead ( https://github.com/servo/ipc-channel/issues/83 ), for someone else to take care of.

PR #80, opened 2016-06-09
linux: Replace some FFI definitions with libc imports

kamalmarhubi:


kamalmarhubi:
`SOCK_SEQPACKET` will be able to go at the next libc release, too.

kamalmarhubi:
libc 0.2.12 is out so I bumped it and dropped `SOCK_SEQPACKET`.

emilio:
This looks good to me, thanks for doing this, and sorry for the extreme delay in reviewing this.

Could you remove all the explicit pointer casts for `as *mut _`/`as *const _` when possible? If for some reason you can't, let's merge it and I'll do it as a followup.

Thanks (and sorry) again! :)

kamalmarhubi:
@emilio no worries about the delay!

Re casts, I don't think any of them can be removed. A cast is needed to go from `&T` to `*T`, and then another from `*T` to `*c_void`. I tried removing a couple and it complained.

If you do remove them as a followup, please ping me in the PR so I can see what you did!

emilio:
@kamalmarhubi ok, fair enough :)

Thanks again!

@bors-servo: r+

antrik:
@kamalmarhubi is there any fundamental reason not to move the remaining FFI definitions to libc as well?

PR #82, opened 2016-06-29
Introduce an inprocess feature

nox:


jdm:
Note that appveyor was introduced to build the in-process code, so please ensure that that continues to be the case.

platform/mod.rs
```diff
@@ -7,71 +7,20 @@
...
 
-#[cfg(target_os="linux")]
-mod linux;
-#[cfg(target_os="macos")]
-mod macos;
-#[cfg(any(target_os="windows", target_os="android"))]
-mod inprocess;
+pub use self::os::{OsIpcChannel, OsIpcOneShotServer, OsIpcReceiver, OsIpcReceiverSet};
+pub use self::os::{OsIpcSelectionResult, OsIpcSender, OsIpcSharedMemory};
+pub use self::os::{OsOpaqueIpcChannel, channel};
```
antrik: I guess that's a matter of taste -- but I strongly dislike the arbitrary split of the use statements... Is there any kind of style guide regarding this?

nox: That's just the exact same split as we use in Servo, keeping lines not too long and not doing multiline use statements.

antrik: I'm not sure what's wrong with multiline use statements... But I can live with that I guess.

Still, I think it would be vastly preferable to group the types in some logical manner. As it is, it is (subjectively) confusing, and (objectively) a PITA if another entry ever needs to be added...

I guess that's bikeshedding though -- feel free to disregard :-)

nox: There is no "logical manner" that fits all of us, so we chose alphabetical order.

pcwalton:
@bors-servo: r+

Nice! This is a great cleanup.

PR #89, opened 2016-07-20
Make in-process implementation threadsafe.

jdm:


jdm:
Tested locally by forcing the in-process implementation to be used in a mac build of Servo.

nox:
@jdm Would it be useful to add an `inprocess` feature that disable platform-specific implementations if requested?

jdm:
Maybe, but then I expect we would need to modify a lot of Cargo.toml files to make use of it.

antrik:
I agree that the `inprocess` backend simply declaring `Sync` on `OsIpcReceiver` is very fishy, and probably needs to be fixed. I don't think wrapping it to actually make it `Sync` is quite right though: apart from adding measurable overhead that is usually not needed, I believe the receiver becoming `Sync` is fundamentally wrong -- it just shifts the synchronisation problem elsewhere, and makes it more obscure. Note that `mpsc::Receiver` itself is `Send` but `!Sync` (and the `RefCell` preserves this), which is an important property of the "single consumer" aspect. Actually using the receiver concurrently from multiple places is bound to introduce race conditions, unless the users make sure all messages are logically independent, or prevent mixups by some kind of synchronisation at a higher level -- in which cases the wrapping should be done individually by these users...

Note that in the Mach (`macos`) backend, `OsIpcReceiver` is indeed `Send` and `!Sync` AIUI. (It is an integer wrapped in a `Cell`.)

(The `linux` variant is technically `Sync` I think, which is just an accident however, because of the socket FD currently simply being a plain integer -- explicitly declaring it `!Sync` doesn't appear to break anything. Along with various other problems, this is yet another hint making me lean more and more towards wrapping this one in a `Cell` as well...)

jdm:
If it's possible to remove the bits that make it Sync then we should.

antrik:
@nox a feature to force use of the `inprocess` backend would be very useful for testing -- I have to work with ugly hacks right now.

Note that this will require some adaptations to the test suite too, as presently some tests make assumptions about the backend in use based on what system we are on.

nox:
@antrik Your wish is my command, my liege. https://github.com/servo/ipc-channel/pull/92

PR #90, opened 2016-08-01
Fix IpcOneShotServer for `inprocess` backend

antrik:
Several patches covering individual aspects, each fixing one possible situation that was broken.

(I don't think there was actually *any* situation that was working correctly before...)

### DIFF
--- a/src/platform/inprocess/mod.rs
+++ b/src/platform/inprocess/mod.rs
@@ -139,7 +139,7 @@ impl OsIpcSender {
     }
 
     pub fn connect(name: String) -> Result<OsIpcSender,MpscError> {
-        let record = ONE_SHOT_SERVERS.lock().unwrap().remove(&name).unwrap();
+        let record = ONE_SHOT_SERVERS.lock().unwrap().get(&name).unwrap().clone();
         record.connect();
         Ok(record.sender)
     }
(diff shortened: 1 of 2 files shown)


antrik:
I tested this by forcing the `inprocess` backend on GNU/Linux -- let's see whether it actually works on Windows as well...

jdm:
```
---- platform::test::server_no_accept stdout ----

	thread 'platform::test::server_no_accept' panicked at 'assertion failed: pingback_rx.try_recv().is_err()', platform/test.rs:426

failures:

    platform::test::server_no_accept
```

jdm:
Looks like the test failure happens on mac.

antrik:
Actually, I am surprised that it didn't fail on the GNU/Linux builder: it turns out this test case fails on my system when using the Linux backend. (I forgot to try that before...)

It seems I made a wrong assumption about how `connect()` behaves on sockets, and consequently how it's supposed to behave in `ipc-channel`. I'll rework the series to account for that.

antrik:
Looks better now.

I also implemented an alternate approach, using a `Drop` trait for cleanup: https://github.com/antrik/ipc-channel/commit/3d1e1401f908c805fee0c7b5b5fbae033a6e2ca0 I'd say it is more elegant -- but I'm not entirely sure it's a good idea, because of the locking involved... Opinions?

pcwalton:
Looks good once conflicts are resolved. Sorry about the delay!

emilio:
Rebased in #100,

Thanks, @antrik! :)

PR #92, opened 2016-08-02
Introduce force-inprocess feature

nox:


antrik:
What's up with this? Does the test failure point to an actual problem, or was that just some kind of Travis hiccup?...

nox:
@antrik Sounds like an actual failure:

```
fatal runtime error: out of memory
fatal runtime error: out of memory
fatal runtime error: out of memory
fatal runtime error: out of memory
fatal runtime error: out of memory
fatal runtime error: out of memory
fatal runtime error: out of memory
fatal runtime error: out of memory
```

Will rebase and retry.

antrik:
@nox oh, I see what happened there: the condition for the fragmenting tests at https://github.com/servo/ipc-channel/pull/92/files#diff-16964b971f706da45fc2c157621dced0R207 needs to be adapted too, so these tests only run when actually using the `linux` backend. They make no sense -- and won't work -- with backends not using fragmentation.

BTW, I wonder whether there is some way to determine the backend to use in a single place, and store this in some kind of config variable that is used throughout the library, rather than having system-specific conditionals everywhere? (If there isn't, I'd consider that a serious shortcoming in Rust/Cargo...)

nox:
@antrik Nice debugging, will disable these tests.

src/test.rs
```diff
@@ -130,8 +129,7 @@ fn select() {
 }
 
 #[test]
-///XXXjdm Windows' libc doesn't include fork.
```
Manishearth: Keep the comment? r=me either way

nox:
@antrik Not running that test wasn't enough.

src/test.rs
antrik: I think keeping the comment would be confusing: the multi-process tests actually can't work / don't make sense on platforms using the `inprocess` implementation -- regardless of whether `fork()` is available or not.

src/platform/test.rs
```diff
@@ -173,41 +173,40 @@ fn big_data_with_sender_transfer() {
...
-                                                    .map(|(tx, rx)| (OsIpcChannel::Sender(tx), rx))
-                                                    .unzip();
-    let (super_tx, super_rx) = platform::channel().unwrap();
-
-    let data: Vec<u8> = (0..size).map(|i| (i % 251) as u8).collect();
-    super_tx.send(&data[..], sender_fds, vec![]).unwrap();
-    let (mut received_data, received_channels, received_shared_memory_regions) =
-        super_rx.recv().unwrap();
+// These tests only apply to platforms that need fragmentation.
+#[cfg(all(not(feature = "inprocess"), target_os = "linux"))]
```
antrik: It's `force-inprocess`, not just `inprocess`.

src/platform/test.rs
```diff
@@ -173,41 +173,40 @@ fn big_data_with_sender_transfer() {
...
+mod fragment_tests {
+    use platform::{self, OsIpcChannel};
 
-    received_data.truncate(size);
-    assert_eq!(received_data.len(), data.len());
-    assert_eq!(&received_data[..], &data[..]);
-    assert_eq!(received_channels.len(), receivers.len());
-    assert_eq!(received_shared_memory_regions.len(), 0);
+    fn with_n_fds(n: usize, size: usize) {
```
antrik: I actually left the function outside of the conditional block on purpose, as it's not really platform-specific, and some existing and/or newly added generic tests could start using it too... I don't feel strongly about this though.

src/platform/test.rs
nox: Oh, I'm silly, thanks.

antrik:
Looks good :-) As you added the warning clean-ups, I guess it needs a new r+? Or is the previous one sufficient to cover this?...

nox:
Just wanted you to take a last look. :)

@bors-servo r=antrik

PR #94, opened 2016-08-07
Linux OsIpcReceiverSet - Switch to use mio

danlrobertson:
Switch from using poll to a level-triggered use of epoll for the Linux OSIpcReceiverSet. The use of epoll should perform better when ther are a large number of watched fd's.

Side note: An edge-triggered use of `epoll` would be very easy here, but to make the transition easier I chose level-triggered. Let me know if you thing edge-triggered would be better. Comments and critiques are welcome!

---

This change is [Reviewable](https://reviewable.io/reviews/servo/ipc-channel/94)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,5 +19,8 @@ serde = "0.8"
 uuid = {version = "0.3", features = ["v4"]}
 fnv = "1.0.3"
 
+[target.'cfg(any(target_os = "linux", target_os = "freebsd"))'.dependencies]
+mio = "0.6.1"
+
 [dev-dependencies]
 crossbeam = "0.2"
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -172,3 +172,143 @@ fn size_22_4m(b: &mut test::Bencher) {
 fn size_23_8m(b: &mut test::Bencher) {
     bench_size(b, 8 * 1024 * 1024);
 }
+
+mod receiver_set {
+    use ipc_channel::ipc::{self, IpcReceiverSet};
+    use test;
+
+    fn gen_select_test(b: &mut test::Bencher, to_send: usize, n: usize) -> () {
+        let mut active = Vec::with_capacity(to_send);
+        let mut dormant = Vec::with_capacity(n - to_send);
+        let mut rx_set = IpcReceiverSet::new().unwrap();
+        for _ in 0..to_send {
+            let (tx, rx) = ipc::channel().unwrap();
+            rx_set.add(rx).unwrap();
+            active.push(tx);
+        }
+        for _ in to_send..n {
+            let (tx, rx) = ipc::channel::<()>().unwrap();
+            rx_set.add(rx).unwrap();
+            dormant.push(tx);
+        }
+        b.iter(|| {
+            for tx in active.iter() {
+                tx.send(()).unwrap();
+            }
+            let mut received = 0;
+            while received < to_send {
+                for result in rx_set.select().unwrap().into_iter() {
+                    let (_, _) = result.unwrap();
+                    received += 1;
+                }
+            }
+        });
+    }
+
+    fn create_empty_set() -> Result<IpcReceiverSet, ()> {
+        Ok(IpcReceiverSet::new().unwrap())
+    }
+
+    fn add_n_rxs(rx_set: &mut IpcReceiverSet, n: usize) -> () {
+        for _ in 0..n {
+            let (_, rx) = ipc::channel::<()>().unwrap();
+            rx_set.add(rx).unwrap();
+        }
+    }
+
+    #[bench]
+    fn send_on_1_of_1(b: &mut test::Bencher) -> () {
+        gen_select_test(b, 1, 1);
+    }
+
+    #[bench]
+    fn send_on_1_of_5(b: &mut test::Bencher) -> () {
+        gen_select_test(b, 1, 5);
+    }
+
+    #[bench]
+    fn send_on_2_of_5(b: &mut test::Bencher) -> () {
+        gen_select_test(b, 2, 5);
+    }
+
+    #[bench]
+    fn send_on_5_of_5(b: &mut test::Bencher) -> () {
+        gen_select_test(b, 5, 5);
+    }
+
+    #[bench]
+    fn send_on_1_of_20(b: &mut test::Bencher) -> () {
+        gen_select_test(b, 1, 20);
+    }
+
+    #[bench]
+    fn send_on_5_of_20(b: &mut test::Bencher) -> () {
+        gen_select_test(b, 5, 20);
+    }
+
+    #[bench]
+    fn send_on_20_of_20(b: &mut test::Bencher) -> () {
+        gen_select_test(b, 20, 20);
+    }
+
+    #[bench]
+    fn send_on_1_of_100(b: &mut test::Bencher) -> () {
+        gen_select_test(b, 1, 100);
+    }
+
+    #[bench]
+    fn send_on_5_of_100(b: &mut test::Bencher) -> () {
+        gen_select_test(b, 5, 100);
+    }
+    #[bench]
+    fn send_on_20_of_100(b: &mut test::Bencher) -> () {
+        gen_select_test(b, 20, 100);
+    }
+
+    #[bench]
+    fn send_on_100_of_100(b: &mut test::Bencher) -> () {
+        gen_select_test(b, 100, 100);
+    }
+
+    #[bench]
+    fn create_and_destroy_empty_set(b: &mut test::Bencher) -> () {
+        b.iter(|| {
+            create_empty_set().unwrap();
+        });
+    }
+
+    #[bench]
+    fn create_and_destroy_set_of_10(b: &mut test::Bencher) -> () {
+        b.iter(|| {
+            let mut rx_set = IpcReceiverSet::new().unwrap();
+            add_n_rxs(&mut rx_set, 10);
+        });
+    }
+
+    #[bench]
+    fn create_and_destroy_set_of_5(b: &mut test::Bencher) -> () {
+        b.iter(|| {
+            let mut rx_set = IpcReceiverSet::new().unwrap();
+            add_n_rxs(&mut rx_set, 5);
+        });
+    }
+
+    #[bench]
+    // Benchmark adding and removing closed receivers from the set
+    fn add_and_remove_closed_receivers(b: &mut test::Bencher) -> () {
+        b.iter(|| {
+            let mut rx_set = IpcReceiverSet::new().unwrap();
+            {
+                {
+                    let (_, rx) = ipc::channel::<()>().unwrap();
+                    rx_set.add(rx).unwrap();
+                }
+                // On select Receivers with a "ClosedChannel" event
+                // will be closed
+                rx_set.select().unwrap();
+                let (_, rx) = ipc::channel::<()>().unwrap();
+                rx_set.add(rx).unwrap();
+            }
+        });
+    }
+}
--- a/src/platform/unix/mod.rs
+++ b/src/platform/unix/mod.rs
@@ -9,9 +9,9 @@
 
 use fnv::FnvHasher;
 use bincode::serde::DeserializeError;
-use libc::{self, MAP_FAILED, MAP_SHARED, POLLIN, PROT_READ, PROT_WRITE, SOCK_SEQPACKET, SOL_SOCKET};
+use libc::{self, MAP_FAILED, MAP_SHARED, PROT_READ, PROT_WRITE, SOCK_SEQPACKET, SOL_SOCKET};
 use libc::{SO_LINGER, S_IFMT, S_IFSOCK, c_char, c_int, c_void, getsockopt};
-use libc::{iovec, mkstemp, mode_t, msghdr, nfds_t, off_t, poll, pollfd, recvmsg, sendmsg};
+use libc::{iovec, mkstemp, mode_t, msghdr, off_t, recvmsg, sendmsg};
 use libc::{setsockopt, size_t, sockaddr, sockaddr_un, socketpair, socklen_t, sa_family_t};
 use std::cell::Cell;
 use std::cmp;
@@ -27,6 +27,8 @@ use std::ptr;
 use std::slice;
 use std::sync::Arc;
 use std::thread;
+use mio::unix::EventedFd;
+use mio::{Poll, Token, Events, Ready, PollOpt};
 
 const MAX_FDS_IN_CMSG: u32 = 64;
 
@@ -77,6 +79,12 @@ pub fn channel() -> Result<(OsIpcSender, OsIpcReceiver),UnixError> {
             Err(UnixError::last())
         }
     }
+}
+
+#[derive(Clone, Copy)]
+struct PollEntry {
+    pub id: u64,
+    pub fd: c_int
 }
 
 #[derive(PartialEq, Debug)]
@@ -415,17 +423,18 @@ impl OsIpcChannel {
 
 pub struct OsIpcReceiverSet {
     incrementor: Incrementor,
-    pollfds: Vec<pollfd>,
-    fdids: HashMap<c_int, u64, BuildHasherDefault<FnvHasher>>
+    poll: Poll,
+    pollfds: HashMap<Token, PollEntry, BuildHasherDefault<FnvHasher>>,
+    events: Events
 }
 
 impl Drop for OsIpcReceiverSet {
     fn drop(&mut self) {
-        unsafe {
-            for pollfd in self.pollfds.iter() {
-                let result = libc::close(pollfd.fd);
-                assert!(thread::panicking() || result == 0);
-            }
+        for &PollEntry { id: _, fd } in self.pollfds.values() {
+            let result = unsafe {
+                libc::close(fd)
+            };
+            assert!(thread::panicking() || result == 0);
         }
     }
 }
@@ -435,60 +444,66 @@ impl OsIpcReceiverSet {
         let fnv = BuildHasherDefault::<FnvHasher>::default();
         Ok(OsIpcReceiverSet {
             incrementor: Incrementor::new(),
-            pollfds: Vec::new(),
-            fdids: HashMap::with_hasher(fnv)
+            poll: try!(Poll::new()),
+            pollfds: HashMap::with_hasher(fnv),
+            events: Events::with_capacity(10)
         })
     }
 
     pub fn add(&mut self, receiver: OsIpcReceiver) -> Result<u64,UnixError> {
         let last_index = self.incrementor.increment();
         let fd = receiver.consume_fd();
-        self.pollfds.push(pollfd {
-            fd: fd,
-            events: POLLIN,
-            revents: 0,
-        });
-        self.fdids.insert(fd, last_index);
+        let io = EventedFd(&fd);
+        let fd_token = Token(fd as usize);
+        let poll_entry = PollEntry {
+            id: last_index,
+            fd: fd
+        };
+        try!(self.poll.register(&io,
+                                fd_token,
+                                Ready::readable(),
+                                PollOpt::level()));
+        self.pollfds.insert(fd_token, poll_entry);
         Ok(last_index)
     }
 
     pub fn select(&mut self) -> Result<Vec<OsIpcSelectionResult>,UnixError> {
         let mut selection_results = Vec::new();
-        let result = unsafe {
-            poll(self.pollfds.as_mut_ptr(), self.pollfds.len() as nfds_t, -1)
+        match self.poll.poll(&mut self.events, None) {
+            Ok(sz) if sz > 0 => {},
+            _ => { return Err(UnixError::last()); }
         };
-        if result <= 0 {
-            return Err(UnixError::last())
-        }
 
-        for pollfd in self.pollfds.iter_mut() {
-            if (pollfd.revents & POLLIN) != 0 {
-                match recv(pollfd.fd, BlockingMode::Blocking) {
-                    Ok((data, channels, shared_memory_regions)) => {
-                        selection_results.push(OsIpcSelectionResult::DataReceived(
-                                *self.fdids.get(&pollfd.fd).unwrap(),
-                                data,
-                                channels,
-                                shared_memory_regions));
-                    }
-                    Err(err) if err.channel_is_closed() => {
-                        let id = self.fdids.remove(&pollfd.fd).unwrap();
-                        unsafe {
-                            libc::close(pollfd.fd);
+        for evt in self.events.iter() {
+            let evt_token = evt.token();
+            match (evt.kind().is_readable(), self.pollfds.get(&evt_token)) {
+                (true, Some(&poll_entry)) => {
+                    match recv(poll_entry.fd, BlockingMode::Blocking) {
+                        Ok((data, channels, shared_memory_regions)) => {
+                            selection_results.push(OsIpcSelectionResult::DataReceived(
+                                    poll_entry.id,
+                                    data,
+                                    channels,
+                                    shared_memory_regions));
                         }
-                        selection_results.push(OsIpcSelectionResult::ChannelClosed(id))
+                        Err(err) if err.channel_is_closed() => {
+                            self.pollfds.remove(&evt_token).unwrap();
+                            unsafe {
+                                libc::close(poll_entry.fd);
+                            }
+                            selection_results.push(OsIpcSelectionResult::ChannelClosed(poll_entry.id))
+                        }
+                        Err(err) => return Err(err),
                     }
-                    Err(err) => return Err(err),
+                },
+                (true, None) => {
+                    panic!("Readable event for unknown token: {:?}", evt_token)
+                },
+                (false, _) => {
+                    panic!("Received an event that was not readable for token: {:?}", evt_token)
                 }
-                pollfd.revents = pollfd.revents & !POLLIN
             }
         }
-
-        // File descriptors not in fdids are closed channels, and the descriptor
-        // has been closed. This must be done after we have finished iterating over
-        // the pollfds vec.
-        let fdids = &self.fdids;
-        self.pollfds.retain(|pollfd| fdids.contains_key(&pollfd.fd));
 
         Ok(selection_results)
     }
@@ -749,6 +764,12 @@ impl From<UnixError> for DeserializeError {
 impl From<UnixError> for Error {
     fn from(unix_error: UnixError) -> Error {
         Error::from_raw_os_error(unix_error.0)
+    }
+}
+
+impl From<Error> for UnixError {
+    fn from(e: Error) -> UnixError {
+        UnixError(e.raw_os_error().unwrap())
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -22,6 +22,9 @@ extern crate serde;
 extern crate uuid;
 #[cfg(all(not(feature = "force-inprocess"), any(target_os = "linux",
                                                 target_os = "freebsd")))]
+extern crate mio;
+#[cfg(all(not(feature = "force-inprocess"), any(target_os = "linux",
+                                                target_os = "freebsd")))]
 extern crate fnv;
 
 pub mod ipc;


danlrobertson:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/platform/linux/mod.rs, line 418 \[r1\]](https://reviewable.io:443/reviews/servo/ipc-channel/94#-KObOVAHs6hbJ6_iSF3A:-KObOVAHs6hbJ6_iSF3B:b8v8wzd) ([raw file](https://github.com/servo/ipc-channel/blob/6ed6eb65da19f713267dc3f5bf91467ab6186445/src/platform/linux/mod.rs#L418)):*
> ```Rust
>     pub fn add(&mut self, receiver: OsIpcReceiver) -> Result<i64,UnixError> {
>         let fd = receiver.consume_fd();
>         self.allfds.push(fd);
> ```

Wasn't a huge fan of having to track all of the fd's being watched, but AFAIK since `OsIpcReceiver` `Drop`s at the end of this function, we need to `dup` the fd.

---

*[src/platform/linux/mod.rs, line 435 \[r1\]](https://reviewable.io:443/reviews/servo/ipc-channel/94#-KObOrTJWQdGN67sXaO4:-KObOrTJWQdGN67sXaO5:b9wa6lt) ([raw file](https://github.com/servo/ipc-channel/blob/6ed6eb65da19f713267dc3f5bf91467ab6186445/src/platform/linux/mod.rs#L435)):*
> ```Rust
>         };
>         let nfds = unsafe {
>             epoll_wait(self.epollfd, events.as_mut_ptr(), MAX_EVENTS, -1) as usize
> ```

It would be reasonably easy to switch from using a slice to a dynamically sized vector. I'm not a huge fan of having a `MAX_EVENTS` chunk size for dealing with events.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/ipc-channel/94)*

antrik:
Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[src/platform/linux/mod.rs, line 418 \[r1\]](https://reviewable.io:443/reviews/servo/ipc-channel/94#-KObOVAHs6hbJ6_iSF3A:-KOg-UEOtIMLV69Vu2fI:b-e2zph2) ([raw file](https://github.com/servo/ipc-channel/blob/6ed6eb65da19f713267dc3f5bf91467ab6186445/src/platform/linux/mod.rs#L418)):*
<details><summary><i>Previously, danlrobertson (Dan Robertson) wrote…</i></summary>
> Wasn't a huge fan of having to track all of the fd's being watched, but AFAIK since `OsIpcReceiver` `Drop`s at the end of this function, we need to `dup` the fd.
</details>
Well, strictly speaking, we do not actually have to `dup()` it; we could prevent the drop instead... However, the actual tracking is necessary, as semantically, the FD is moved into the select structure.

I actually have a pending PR for avoiding the unnecessary `dup()` here (and for a few similar cases); but I decided to postpone it in favour of investigating the possibility of introducing a wrapper type for FDs, which should enable handling these situations in a more robust and natural fashion...

To keep changes minimal, I'd say it's fine for your PR to preserve the existing approach.

---

*[src/platform/linux/mod.rs, line 435 \[r1\]](https://reviewable.io:443/reviews/servo/ipc-channel/94#-KObOrTJWQdGN67sXaO4:-KOg1ZPUNBvqcKmHT9dW:b-57g9xt) ([raw file](https://github.com/servo/ipc-channel/blob/6ed6eb65da19f713267dc3f5bf91467ab6186445/src/platform/linux/mod.rs#L435)):*
<details><summary><i>Previously, danlrobertson (Dan Robertson) wrote…</i></summary>
> It would be reasonably easy to switch from using a slice to a dynamically sized vector. I'm not a huge fan of having a `MAX_EVENTS` chunk size for dealing with events.
</details>
I don't see how you want to avoid the chunking here: I believe it's inherent in the `epoll` API?... (To make it more transparent, I guess you *could* implement a loop that tries to catch all events before returning -- but I'm not sure it's worthwhile, or doesn't come with undesired side effects...)

In the name of defensive programming, I would however indeed suggest using a `Vec` here: then doing a `set_len()` after the `epoll_wait()`; and wrapping this entire part in one `unsafe{}` block. The following code can then simply iterate over the `Vec` in a perfectly safe fashion. The way it is now, the unsafety spills out into the surrounding code: it relies upon the "safe" code to handle things correctly in order to avoid undefined behaviour, which is very very bad... (Most notably, if `epoll_event` ever were to grow a `Drop` implementation for some reason, hell would break loose.)

---

*[src/platform/linux/mod.rs, line 399 \[r2\]](https://reviewable.io:443/reviews/servo/ipc-channel/94#-KOg6jzjo04c2YHM8Ezc:-KOg6jzjo04c2YHM8Ezd:b-9on7s) ([raw file](https://github.com/servo/ipc-channel/blob/9719339f3a55dc1796443233d42ca2adc691a76d/src/platform/linux/mod.rs#L399)):*
> ```Rust
>         unsafe {
>             for fd in self.allfds.iter() {
>                 let result = libc::close(*fd);
> ```

I'm still a bit of a noob in Rust, so I hope this isn't a silly question: but why are you explicitly dereferencing `fd` here? I don't think that's standard practice in a situation like this?...

---

*[src/platform/linux/mod.rs, line 461 \[r2\]](https://reviewable.io:443/reviews/servo/ipc-channel/94#-KOg4a7DrXoiKRJHxfwv:-KOg4a7DrXoiKRJHxfww:b6wpt5t) ([raw file](https://github.com/servo/ipc-channel/blob/9719339f3a55dc1796443233d42ca2adc691a76d/src/platform/linux/mod.rs#L461)):*
> ```Rust
>                     Err(err) => return Err(err),
>                 }
>             } else if evt.events & (EPOLLERR | EPOLLHUP) as u32 > 0 {
> ```

I'm not sure it's actually correct to handle this here; and in my understanding, the behaviour doesn't differ between `poll()` and `epoll` -- so this is actually an unrelated change. I suggest putting it in a separate PR, along with new test cases to cover these situations.

---

*[src/platform/linux/mod.rs, line 474 \[r2\]](https://reviewable.io:443/reviews/servo/ipc-channel/94#-KOg5NnreM3m9_Dnss0o:-KOg5NnreM3m9_Dnss0p:b-ulv3n5) ([raw file](https://github.com/servo/ipc-channel/blob/9719339f3a55dc1796443233d42ca2adc691a76d/src/platform/linux/mod.rs#L474)):*
> ```Rust
>                     libc::close(*hangup);
>                 }
>             }
> ```

I don't understand why you added this here: am I missing some other change that makes it necessary? Or do you think this is a bug in the original implementation? (In which case I'd suggest a separate PR, or at the very least a separate commit...)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/ipc-channel/94)*

antrik:
I'd like to point out that although the platform backend is named `linux`, thus far it has really just been a generic UNIX implementation, which should essentially work on any POSIX system. (Minus some small wrinkles probably...) Switching to an actually Linux-specific interface will make it more painful to port to other systems later on -- I think we first need to make plans how to handle this going forward...

I wonder whether there is something specifically motivating this change; or it just looked like it might be a good idea in general?

Also, when doing performance optimisation, I suggest adding benchmark tests to keep track of actual results...

danlrobertson:
Great points. So actually the reason why I opened this PR, was because I started porting this to FreeBSD for servo/servo#11625 and was going to use `kqueue`, and noticed `linux` didn't use `epoll`, and I figured that would be a usefull change as well. There enough differences (e.g. `sockaddr_un` and a few other type differences) that some refractoring was going to need to happen anyways. As a whole, I'd have to say it isn't necessary though... Just thought it might be a good idea.

---

Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[src/platform/linux/mod.rs, line 418 \[r1\]](https://reviewable.io:443/reviews/servo/ipc-channel/94#-KObOVAHs6hbJ6_iSF3A:-KOgMgmMxBFJkKoGbG56:bylkcz2) ([raw file](https://github.com/servo/ipc-channel/blob/6ed6eb65da19f713267dc3f5bf91467ab6186445/src/platform/linux/mod.rs#L418)):*
<details><summary><i>Previously, antrik (Olaf Buddenhagen) wrote…</i></summary>
> Well, strictly speaking, we do not actually have to `dup()` it; we could prevent the drop instead... However, the actual tracking is necessary, as semantically, the FD is moved into the select structure.
> 
> I actually have a pending PR for avoiding the unnecessary `dup()` here (and for a few similar cases); but I decided to postpone it in favour of investigating the possibility of introducing a wrapper type for FDs, which should enable handling these situations in a more robust and natural fashion...
> 
> To keep changes minimal, I'd say it's fine for your PR to preserve the existing approach.
</details>
Awesome! That would be really great

---

*[src/platform/linux/mod.rs, line 435 \[r1\]](https://reviewable.io:443/reviews/servo/ipc-channel/94#-KObOrTJWQdGN67sXaO4:-KOgEXaD-vgJN8uWXWDC:b3g8pny) ([raw file](https://github.com/servo/ipc-channel/blob/6ed6eb65da19f713267dc3f5bf91467ab6186445/src/platform/linux/mod.rs#L435)):*
<details><summary><i>Previously, antrik (Olaf Buddenhagen) wrote…</i></summary>
> I don't see how you want to avoid the chunking here: I believe it's inherent in the `epoll` API?... (To make it more transparent, I guess you *could* implement a loop that tries to catch all events before returning -- but I'm not sure it's worthwhile, or doesn't come with undesired side effects...)
> 
> In the name of defensive programming, I would however indeed suggest using a `Vec` here: then doing a `set_len()` after the `epoll_wait()`; and wrapping this entire part in one `unsafe{}` block. The following code can then simply iterate over the `Vec` in a perfectly safe fashion. The way it is now, the unsafety spills out into the surrounding code: it relies upon the "safe" code to handle things correctly in order to avoid undefined behaviour, which is very very bad... (Most notably, if `epoll_event` ever were to grow a `Drop` implementation for some reason, hell would break loose.)
</details>
So you do need some sort of chunking, but the chunking can be more "configurable". There are a few ways this could work. But they're probably not worthwhile (especially no. 2)

1)  As opposed to using `MAX_EVENTS` you would have some configurable size `x` in the constructor, so then as opposed to creating a slice you create a vector of size `x` and then use `epoll_wait(epollfd, vec.as_mut_ptr(), vec.len())`

2) Use penalties. Again use a config value `x`. If `nfds` is greater than or equal to `threashold(x)` increase `x` with `MIN(MAXIMUM_ALLOWED, ++x)` or if it is less than `threashold(x)` decrease it with `MAX(MINIMUM_ALLOWED, --x`.

Ah! Good point. Thanks for the tip!

---

*[src/platform/linux/mod.rs, line 399 \[r2\]](https://reviewable.io:443/reviews/servo/ipc-channel/94#-KOg6jzjo04c2YHM8Ezc:-KOgDcEd1LgTN1YA9p0J:b-uunakp) ([raw file](https://github.com/servo/ipc-channel/blob/9719339f3a55dc1796443233d42ca2adc691a76d/src/platform/linux/mod.rs#L399)):*
<details><summary><i>Previously, antrik (Olaf Buddenhagen) wrote…</i></summary>
> I'm still a bit of a noob in Rust, so I hope this isn't a silly question: but why are you explicitly dereferencing `fd` here? I don't think that's standard practice in a situation like this?...
</details>
Good question... I'm a bit of a Rust noob as well. `allfds` can't be moved out of the borrowed context, and as a result fd is a `&i32`. While `libc::close` requires a standard int

---

*[src/platform/linux/mod.rs, line 461 \[r2\]](https://reviewable.io:443/reviews/servo/ipc-channel/94#-KOg4a7DrXoiKRJHxfwv:-KOgK0UwfDiWJFP_IBfm:b-z11f4r) ([raw file](https://github.com/servo/ipc-channel/blob/9719339f3a55dc1796443233d42ca2adc691a76d/src/platform/linux/mod.rs#L461)):*
<details><summary><i>Previously, antrik (Olaf Buddenhagen) wrote…</i></summary>
> I'm not sure it's actually correct to handle this here; and in my understanding, the behaviour doesn't differ between `poll()` and `epoll` -- so this is actually an unrelated change. I suggest putting it in a separate PR, along with new test cases to cover these situations.
</details>
Np. I'll remove it and submit a separate PR if necessary.

---

*[src/platform/linux/mod.rs, line 474 \[r2\]](https://reviewable.io:443/reviews/servo/ipc-channel/94#-KOg5NnreM3m9_Dnss0o:-KOgKyOz5IUP-hzmrhOD:b-kdt0rl) ([raw file](https://github.com/servo/ipc-channel/blob/9719339f3a55dc1796443233d42ca2adc691a76d/src/platform/linux/mod.rs#L474)):*
<details><summary><i>Previously, antrik (Olaf Buddenhagen) wrote…</i></summary>
> I don't understand why you added this here: am I missing some other change that makes it necessary? Or do you think this is a bug in the original implementation? (In which case I'd suggest a separate PR, or at the very least a separate commit...)
</details>
Ah, actually the `libc::close` is a mistake (I removed the wrong line last night... I'm pretty sure that's why the build is failing?), but the `epoll_ctl` is necessary, so that we stop watching the fd after it hits a hangup. Let me know if I've misunderstood something

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/ipc-channel/94#-:-KOgMmnQwmZXTUJAylOS:bnidf4w)*

danlrobertson:
Thanks for the comments! If you disagree with the `kqueue`/`epoll` approach or have more questions please let me know.

---

Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/ipc-channel/94#-:-KOgPHUDgYL-jOOej3UQ:bdfv04y)*

antrik:
Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[src/platform/linux/mod.rs, line 435 \[r1\]](https://reviewable.io:443/reviews/servo/ipc-channel/94#-KObOrTJWQdGN67sXaO4:-KOgZ3GIXJ5aiGEMDjcL:b-g4im4) ([raw file](https://github.com/servo/ipc-channel/blob/6ed6eb65da19f713267dc3f5bf91467ab6186445/src/platform/linux/mod.rs#L435)):*
<details><summary><i>Previously, danlrobertson (Dan Robertson) wrote…</i></summary>
> So you do need some sort of chunking, but the chunking can be more "configurable". There are a few ways this could work. But they're probably not worthwhile (especially no. 2)
> 
> 1)  As opposed to using `MAX_EVENTS` you would have some configurable size `x` in the constructor, so then as opposed to creating a slice you create a vector of size `x` and then use `epoll_wait(epollfd, vec.as_mut_ptr(), vec.len())`
> 
> 2) Use penalties. Again use a config value `x`. If `nfds` is greater than or equal to `threashold(x)` increase `x` with `MIN(MAXIMUM_ALLOWED, ++x)` or if it is less than `threashold(x)` decrease it with `MAX(MINIMUM_ALLOWED, --x`.
> 
> Ah! Good point. Thanks for the tip!
</details>
I'm just not convinced it's worthwhile. It might indeed be perfectly fine to go with `MAX_EVENTS` set to 1... This would in fact be more intuitive for something called `select()` I'd say -- I suspect it just behaves differently in the original implementation because `poll()` kinda forces this?...

I just realised though that the select won't yield any actual results (I think) if all the events we get from `epoll_wait()` are of a "wrong" type... Which is mostly likely true of the `poll()` implementation as well?... So many intricate details to consider :-(

---

*[src/platform/linux/mod.rs, line 399 \[r2\]](https://reviewable.io:443/reviews/servo/ipc-channel/94#-KOg6jzjo04c2YHM8Ezc:-KOgXclz7-AWUbG_iNbX:b-xrmmve) ([raw file](https://github.com/servo/ipc-channel/blob/9719339f3a55dc1796443233d42ca2adc691a76d/src/platform/linux/mod.rs#L399)):*
<details><summary><i>Previously, danlrobertson (Dan Robertson) wrote…</i></summary>
> Good question... I'm a bit of a Rust noob as well. `allfds` can't be moved out of the borrowed context, and as a result fd is a `&i32`. While `libc::close` requires a standard int
</details>
Well, I am aware that the iterator returns `&i32` values -- it's just that since this is a function argument, it would be handled transparently by an automatic deref coercion. (Unless I am very confused...) And as far as I am aware, we do not generally spell out dereferencing if it's covered by coercion?

---

*[src/platform/linux/mod.rs, line 474 \[r2\]](https://reviewable.io:443/reviews/servo/ipc-channel/94#-KOg5NnreM3m9_Dnss0o:-KOgaG6Xalmm3vVhegfT:bob1yg3) ([raw file](https://github.com/servo/ipc-channel/blob/9719339f3a55dc1796443233d42ca2adc691a76d/src/platform/linux/mod.rs#L474)):*
<details><summary><i>Previously, danlrobertson (Dan Robertson) wrote…</i></summary>
> Ah, actually the `libc::close` is a mistake (I removed the wrong line last night... I'm pretty sure that's why the build is failing?), but the `epoll_ctl` is necessary, so that we stop watching the fd after it hits a hangup. Let me know if I've misunderstood something
</details>
Yeah, that sounds right -- an `epoll_ctl()` here would totally make sense.

(Closing the descriptor actually removes it from the set automatically AIUI -- but only if there are no other outstanding FDs referencing the same underlying `open()`; and also, it would potentially discard pending messages, which we don't want...)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/ipc-channel/94)*

antrik:
I do not exactly disagree with the `epoll`/`kqueue` approach -- I just wonder whether the extra complexity is justified here...

More importantly, I think it is really important to find a sane approach for handling such minor variations in general. I can't speak for @pcwalton and others -- but I for my part would be very sad if we end up with a `FreeBSD` backend that is just a copy of the `Linux` one with only some small changes, and another one for generic UNIX...

(Or a `Hurd` backend that is mostly a copy of `MacOS` one, for that matter.)

notriddle:
Call me insane, but why can't we just use [mio with UNIX sockets](http://rustdoc.s3-website-us-east-1.amazonaws.com/mio/master/mio/unix/index.html)?

antrik:
@notriddle can you elaborate please? Where exactly would `mio` come in?

notriddle:
mio provides an abstraction layer between async IO apis, such as epoll and kqueue.

antrik:
@notriddle that *might* work I guess... (Though the `Poll` struct actually seems more relevant than the `unix` module for this.)

Can't tell for sure, as most of `mio` doesn't appear to have actual documentation?...

danlrobertson:
After taking a quick peek at `mio` I aggree.

The `Poll` struct seems quite relevant, but to be honest. I don't think the `poll` portion is going to be the hardest part in the road to supporting more OSes. For example, this is my current progress on moving ipc-channel to freebsd https://github.com/danlrobertson/ipc-channel/tree/support-freebsd. The current implementation of `select` with poll works, but a huge number of other tests fail.

Take this with a grain of salt, as this is my first PR to ipc.

danlrobertson:
Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/platform/linux/mod.rs, line 435 \[r1\]](https://reviewable.io:443/reviews/servo/ipc-channel/94#-KObOrTJWQdGN67sXaO4:-KOhGRnwmtCYKq397-sy:b-7q9ul) ([raw file](https://github.com/servo/ipc-channel/blob/6ed6eb65da19f713267dc3f5bf91467ab6186445/src/platform/linux/mod.rs#L435)):*
<details><summary><i>Previously, antrik (Olaf Buddenhagen) wrote…</i></summary>
> I'm just not convinced it's worthwhile. It might indeed be perfectly fine to go with `MAX_EVENTS` set to 1... This would in fact be more intuitive for something called `select()` I'd say -- I suspect it just behaves differently in the original implementation because `poll()` kinda forces this?...
> 
> I just realised though that the select won't yield any actual results (I think) if all the events we get from `epoll_wait()` are of a "wrong" type... Which is mostly likely true of the `poll()` implementation as well?... So many intricate details to consider :-(
</details>
I agree. If it is decided to use `epoll/kqueue` a more complex implementation probably isn't worth it.

Ah! yes, that is an astute observation. For both implementations, if none of the results are POLLIN the returned vec will be empty

---

*[src/platform/linux/mod.rs, line 399 \[r2\]](https://reviewable.io:443/reviews/servo/ipc-channel/94#-KOg6jzjo04c2YHM8Ezc:-KOhGHFBBRJcQXIBPt4W:bomeyyb) ([raw file](https://github.com/servo/ipc-channel/blob/9719339f3a55dc1796443233d42ca2adc691a76d/src/platform/linux/mod.rs#L399)):*
<details><summary><i>Previously, antrik (Olaf Buddenhagen) wrote…</i></summary>
> Well, I am aware that the iterator returns `&i32` values -- it's just that since this is a function argument, it would be handled transparently by an automatic deref coercion. (Unless I am very confused...) And as far as I am aware, we do not generally spell out dereferencing if it's covered by coercion?
</details>
I expected that as well, but it would not compile without the Deref

---

*[src/platform/linux/mod.rs, line 474 \[r2\]](https://reviewable.io:443/reviews/servo/ipc-channel/94#-KOg5NnreM3m9_Dnss0o:-KOhHZ9d5vXYH5-QAQHZ:b-uxwia) ([raw file](https://github.com/servo/ipc-channel/blob/9719339f3a55dc1796443233d42ca2adc691a76d/src/platform/linux/mod.rs#L474)):*
<details><summary><i>Previously, antrik (Olaf Buddenhagen) wrote…</i></summary>
> Yeah, that sounds right -- an `epoll_ctl()` here would totally make sense.
> 
> (Closing the descriptor actually removes it from the set automatically AIUI -- but only if there are no other outstanding FDs referencing the same underlying `open()`; and also, it would potentially discard pending messages, which we don't want...)
</details>
Yup. Seems to have fixed the build. Thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/ipc-channel/94)*

antrik:
@danlrobertson well, in general, if there is an existing abstraction we can use, we probably should -- especially if it involves messy things like platform-specific and unsafe code... The problem is really just that from a quick glance at the `mio` "documentation" -- which for the most part is just a cross-reference really -- I can't tell whether it is likely to work or not. But it's certainly worth some investigation...

danlrobertson:
Added some benchmarks for `poll`. The first results I got for the older version and the newer version were.

```
$ git checkout switch_to_epoll
$ cargo bench poll
   Compiling ipc-channel v0.5.1 (file:///home/drobertson/git/servo/ipc-channel)
       Finished release [optimized] target(s) in 68.47 secs
            Running target/release/bench-cd0b4141c4f39269

            running 3 tests
            test poll::bench_fifty_percent       ... bench:     297,148 ns/iter (+/- 1,959)
            test poll::bench_one_hundred_percent ... bench:     594,874 ns/iter (+/- 4,740)
            test poll::bench_one_percent         ... bench:       5,368 ns/iter (+/- 129)

            test result: ok. 0 passed; 0 failed; 0 ignored; 3 measured

$ git checkout HEAD~1
$ cargo bench poll
   Compiling ipc-channel v0.5.1 (file:///home/drobertson/git/servo/ipc-channel)
       Finished release [optimized] target(s) in 8.99 secs
            Running target/release/bench-cd0b4141c4f39269

            running 3 tests
            test poll::bench_fifty_percent       ... bench:     487,272 ns/iter (+/- 3,168)
            test poll::bench_one_hundred_percent ... bench:   1,146,511 ns/iter (+/- 11,361)
            test poll::bench_ten_percent         ... bench:      13,372 ns/iter (+/- 489)

            test result: ok. 0 passed; 0 failed; 0 ignored; 3 measured
```

I repeated it a few times and received similar results. Please double check my benchmark code... This was the first time I've written benchmarks in rust. Also I'm really curious if others get similar results. If you test it out on your boxen, I'd be interested to know your results.

danlrobertson:
I just created a branch [here](https://github.com/danlrobertson/ipc-channel/tree/switch_to_mio) attempting to use `mio` instead. Interestingly, the performance on the benchmark tests I just wrote were almost identical to the implementation with regular `poll`. Note: if we were to use `mio` we would need to use `0.6.x`.

benches/bench.rs
```diff
@@ -170,3 +170,51 @@ fn size_22_4m(b: &mut test::Bencher) {
 fn size_23_8m(b: &mut test::Bencher) {
     bench_size(b, 8 * 1024 * 1024);
 }
+
+macro_rules! gen_poll_test {
```
antrik: This is what I originally did for the other benchmark tests -- but @pcwalton didn't like the use of macros here...

benches/bench.rs
```diff
@@ -170,3 +170,51 @@ fn size_22_4m(b: &mut test::Bencher) {
...
+            b.iter(|| {
+                for tx in active.iter() {
+                    tx.send("Just a flesh wound".to_owned()).unwrap();
+                }
+                let mut received = 0;
+                while received < $to_send {
+                    for result in rx_set.select().unwrap().into_iter() {
+                        let (_, received_data) = result.unwrap();
+                        let received_string: String = received_data.to().unwrap();
+                        assert_eq!(received_string, "Just a flesh wound");
```
antrik: I wouldn't do that in a benchmark test. It might have a pretty significant effect on the result...

antrik: In fact -- as much as I like the quote :-) -- I'd probably use a channel of type `()`, to minimise the overhead of anything but the `select()` itself...

benches/bench.rs
```diff
@@ -170,3 +170,51 @@ fn size_22_4m(b: &mut test::Bencher) {
...
+}
+
+#[cfg(test)]
+mod poll {
+    use ipc_channel::ipc::{self, IpcReceiverSet};
+    use test;
+
+    gen_poll_test!{bench_one_percent, 1, 100}
+    gen_poll_test!{bench_fifty_percent, 50, 100}
+    gen_poll_test!{bench_one_hundred_percent, 100, 100}
```
antrik: I'm sure we could get more meaningful cases here: while the 100% one is useful to check the extreme end, the 50% one really doesn't tell us much. I'd rather go for 2 and 5 or something like that, which are way more likely in the real world I suspect...

Also, tests with smaller sets would be useful -- including the extreme (but not unlikely) case of just a single channel; and maybe 5 and 20 or so for likely real world cases.

(If you are *really* ambitious, you could add some instrumentation, to get an idea of the usage patterns actually coming up in Servo, so we don't have to rely on guesswork... I must admit though that I would only do that myself if I have absolutely no idea what might be realistic :-) )

On a related note, I guess it might be useful to also have some tests for the costs of creating/manipulating/destroying the set...

antrik:
Whoops, forgot that you were using Reviewable here... Hope you don't mind.

I'm not sure what to make of the disappointing `mio` results. Do you have any idea what causes the overhead there? Is it even using `epoll`?...

benches/bench.rs
danlrobertson: Updated to use a function instead

benches/bench.rs
danlrobertson: Good point. I updated the type and it did help. Thanks! the results are more like what I expected. I'll try `mio` again.

benches/bench.rs
danlrobertson: > Also, tests with smaller sets would be useful -- including the extreme (but not unlikely) case of just a single channel; and maybe 5 and 20 or so for likely real world cases.

Ah true. I'll remove the 50% case and make a few more reasonable cases. I wanted to have at least the 1% for when we expect epoll to perform way better than poll and 100% where we expect it to be a bit closer. After I push those changes I'll look into adding bench marks for the destruction of the sets.

danlrobertson:
So after the removal of the assert the results are more in line with what is expected. mio performs better than the original, but not as good as using epoll directly

**original**
```
running 3 tests
test poll::bench_fifty_percent       ... bench:     330,639 ns/iter (+/- 6,048)
test poll::bench_one_hundred_percent ... bench:     747,497 ns/iter (+/- 10,552)
test poll::bench_one_percent         ... bench:      13,407 ns/iter (+/- 216)
```

**mio**
```
running 3 tests
test poll::bench_fifty_percent       ... bench:     305,776 ns/iter (+/- 3,245)
test poll::bench_one_hundred_percent ... bench:     717,213 ns/iter (+/- 8,743)
test poll::bench_one_percent         ... bench:       4,931 ns/iter (+/- 240)
```

**epoll**
```
running 3 tests
test poll::bench_fifty_percent       ... bench:     201,878 ns/iter (+/- 3,569)
test poll::bench_one_hundred_percent ... bench:     403,554 ns/iter (+/- 5,512)
test poll::bench_one_percent         ... bench:       4,855 ns/iter (+/- 134)
```

AFAIK mio will not perform better than directly using epoll for our use case due to the way mio wraps epoll. It does use `kqueue/epoll`, but the overhead of making their wrapper fit our current design is probably why that implementation performs worse. E.g. getting the file descriptor from the event is super easy and efficient when directly using epoll/poll, but dificult and inefficient when using mio (I could have missed an optimization in my implementation, so this could be my fault).

metajack:
Is it possible to optimize this upstream in mio? It's a shame we can't rely on that.

alexcrichton:
@danlrobertson hello! I'd be quite curious to drill into what's happening [with these numbers](https://github.com/servo/ipc-channel/pull/94#issuecomment-241212426), do you have a few branches that I could check out and reproduce those results locally?

alexcrichton:
@danlrobertson I took a brief glance at [your commit to use mio](https://github.com/danlrobertson/ipc-channel/commit/d539e346d83cd423dafff7ae048066d6cb546812) and one thing I noticed was that each time `poll` is called [it calls `Events::new`](https://github.com/danlrobertson/ipc-channel/commit/d539e346d83cd423dafff7ae048066d6cb546812#diff-327c56c243dbf015b41d7718144dd970R414). The `Events` structure is relatively expensive to create and is intended to be cached across calls to `poll`. If that's done do the numbers look better?

carllerche:
@danlrobertson +1 on @alexcrichton's comment. Mio should be zero cost over epoll. If you are not seeing this, there is a bug somewhere.

However, as @alexcrichton pointed out, it is expected that `Events` is created once per `Poll` instance. So, you are expected to reuse the same `Events` value each call to `Poll::poll`. Under the hood, `Events` is backed by a vector and initialized w/ the requested capacity. This is not a cheap operation (relatively speaking).

danlrobertson:
> The Events structure is relatively expensive to create and is intended to be cached across calls to poll.

Ah! good catch. Not sure why I did that.

> Mio should be zero cost over epoll. If you are not seeing this, there is a bug somewhere.

Good to know. So what about the use of a `BTreeMap<Token, Io>` and `get(event.token())` to get the fd associated with the event. Is there a more efficient method of going about this?

carllerche:
@danlrobertson I tend to use `Slab` https://github.com/carllerche/slab. It is like a vector map, but handles managing the slot indexes (tokens) for you.

danlrobertson:
@alexcrichton changing `select` to not call `Events::new` on each call didn't change the performance that much. I'll keep looking at it though.

`mio` performance is equal to directly using `epoll` for smaller sets, but as the set size increases, it starts to diverge. I'll take some time to look at my commit and how I could optimize it. I will also try to take a close look at mio to find possible places that I've misused it and/or features that can be added to make it more usable in this case.

carllerche:
@danlrobertson did you profile? Given that it diverges as the number of IO elements grows that this is related to using `BTreeMap`. I would recommend using `Slab` or even just a `Vec` like your epoll code does.

notriddle:
As a reminder, BTreeMap isn't constant-time. It's n*log(n) to insert and remove from it.

danlrobertson:
Yeah should have mentioned that. I tried using a `HashMap` and `Slab`, but the rest of the code didn't change. If it would help, I can submit a separate PR with the `mio` changes

antrik:
@danlrobertson any progress with this?

The numbers you have presented so far show that the extra overhead is in handling the many events, not in the wait call itself. Seeing some numbers for a smaller set size might allow *some* guesses as to where the overhead is -- but actually profiling the code is probably more useful.

I'm wondering whether the `mio` variant is perhaps only processing a single event at a time, or something like that?... But that's really just guesswork.

Also, I'm still hoping to see numbers for the case of just a few events. If these aren't significantly worse than the manual implementation, I'd consider it good enough for practical purposes.

danlrobertson:
Actually, I'm planning on working on this today. I switched the tests over to include smaller sets, but just haven't run the tests yet.

> I'm wondering whether the mio variant is perhaps only processing a single event at a time, or something like that?... But that's really just guesswork.

I'm actually thinking that this is the case too. Earlier this week I thought of this, so forced it to return one at a time, and the benchmark results were no different.

> If these aren't significantly worse than the manual implementation, I'd consider it good enough for practical purposes.

Sounds good. I should have an update soon.

danlrobertson:
The following are the most recent results.

**using epoll directly**
```
test receiver_set::bench_create_and_destroy_10   ... bench:     126,673 ns/iter (+/- 14,297)
test receiver_set::bench_create_and_destroy_5    ... bench:      46,943 ns/iter (+/- 3,464)
test receiver_set::bench_manipulation            ... bench:      33,992 ns/iter (+/- 2,028)
test receiver_set::bench_one_hundred_percent_100 ... bench:     416,418 ns/iter (+/- 28,142)
test receiver_set::bench_one_percent_100         ... bench:       4,727 ns/iter (+/- 222)
test receiver_set::bench_thirty_percent_10       ... bench:      11,212 ns/iter (+/- 362)
test receiver_set::bench_twenty_percent_10       ... bench:       7,949 ns/iter (+/- 360)
test receiver_set::bench_twenty_percent_5        ... bench:       4,626 ns/iter (+/- 560)
```
**using mio**
```
running 8 tests
test receiver_set::bench_create_and_destroy_10   ... bench:     145,227 ns/iter (+/- 5,101)
test receiver_set::bench_create_and_destroy_5    ... bench:      59,996 ns/iter (+/- 797)
test receiver_set::bench_manipulation            ... bench:      46,405 ns/iter (+/- 356)
test receiver_set::bench_one_hundred_percent_100 ... bench:     735,510 ns/iter (+/- 5,694)
test receiver_set::bench_one_percent_100         ... bench:       4,744 ns/iter (+/- 380)
test receiver_set::bench_thirty_percent_10       ... bench:      11,355 ns/iter (+/- 336)
test receiver_set::bench_twenty_percent_10       ... bench:       8,065 ns/iter (+/- 257)
test receiver_set::bench_twenty_percent_5        ... bench:       4,758 ns/iter (+/- 137)
```

Again, these results might be skewed because I have a good bit of experience using epoll/kqueue and until now no experience with mio :smile:. Any critiques are appreciated. `HashMap` performed much better than `Slab`, so I used `HashMap`. The branch using mio can be found [here](https://github.com/danlrobertson/ipc-channel/tree/switch_to_mio). In the meantime, I'll be running examples under callgrind to try to find possible performance issues.

emilio:
Hmm... Just curious, how does it perform if the HashMap uses Fnv instead
of rust's default hasher?

danlrobertson:
> Just curious, how does it perform if the HashMap uses Fnv instead

Great question! This is is a great case for FNV. There was a small improvement in performance when using this.

carllerche:
@danlrobertson it looks like your branch is using `HashMap` and not `Slab`. If `Slab` is slower than `HashMap` then something is pretty wrong given that `Slab` is just an index lookup into a vector.

danlrobertson:
Yup, that is intended. Slab was slower.

carllerche:
Do you have the code w/ Slab? Because that is quite surprising.

danlrobertson:
Since there seems to be more interest in using `mio`, I've updated the branch to contain the code using `mio`.

@carllerche I'll get you the code using slab. IMO we're starting to drift outside the scope of this PR if we go too much more in depth with that, but if you'd like to discuss it more I can open an Issue on the Slab repo.

danlrobertson:
Woohoo! got the large set bench time down to 755,000 ns/iter!

antrik:
@danlrobertson sorry for dropping the ball on this :-( When I came back from vacation there was so much going on, and it kinda fell through the cracks...

So, looking at the latest benchmark results you posted, the selection of test cases still seems a bit arbitrary to me -- I'd go for something along these lines: 1/1, 1/5, 2/5, 5/5, 1/20, 5/20, 20/20, 1/100, 5/100, 20/100, 100/100. (Using absolute numbers seems less confusing to me than percentages BTW...) I think this would give more clue where the performance differences manifest, and whether they are likely to matter in practice. Right now, the only case having a large enough difference is an extreme case, and there is just not enough data to extrapolate whether it affects realistic use cases too...

I think it still would be useful to post the code variant using `Slab` somewhere, so people could try to figure out why it performs so shockingly bad...

Ironically, this PR got broken by your own PR adding FreeBSD support :-) Shouldn't be hard to fix though...

(And while at it, can you please update the title, to reflect that this is about `mio` now?...)

antrik:
Oh, and the commit message also need an update...

src/platform/linux/mod.rs
```diff
@@ -386,76 +390,84 @@ impl OsIpcChannel {
...
-                                data,
-                                channels,
-                                shared_memory_regions));
-                    }
-                    Err(err) if err.channel_is_closed() => {
-                        hangups.insert(pollfd.fd);
-                        selection_results.push(OsIpcSelectionResult::ChannelClosed(
-                                    pollfd.fd as i64))
+        for i in 0..sz {
+            let evt = self.events.get(i).unwrap();
```
danlrobertson: At the time of writing this I think events didn't impl Iterator? maybe? I know there was some reason why I did this, but yes. Now this is possible.

src/platform/linux/mod.rs
```diff
@@ -386,76 +390,84 @@ impl OsIpcChannel {
...
+            let evt_token = evt.token();
+            match (evt.kind().is_readable(), self.pollfds.get(&evt_token)) {
+                (true, Some(fd)) => {
```
danlrobertson: Ah! Nice! Much more readable.

src/platform/linux/mod.rs
```diff
@@ -386,76 +390,84 @@ impl OsIpcChannel {
...
-                }
-                pollfd.revents = pollfd.revents & !POLLIN
+                },
+                _ => {},
             }
         }
 
         if !hangups.is_empty() {
-            self.pollfds.retain(|pollfd| !hangups.contains(&pollfd.fd));
+            for hangup in hangups.iter() {
```
danlrobertson: Indeed.

src/platform/linux/mod.rs
```diff
@@ -386,76 +390,84 @@ impl OsIpcChannel {
...
+                self.pollfds.remove(hangup).unwrap();
+            }
         }
 
         Ok(selection_results)
     }
 }
 
+impl Drop for OsIpcReceiverSet {
```
danlrobertson: I completely removed `Drop` at some point and when I added it back, I added it here. It is moved in the latest update.

src/platform/linux/mod.rs
```diff
@@ -386,76 +390,84 @@ impl OsIpcChannel {
...
+    fn drop(&mut self) {
+        for pair in &self.pollfds {
```
danlrobertson: Indeed.

src/platform/linux/mod.rs
```diff
@@ -386,76 +390,84 @@ impl OsIpcChannel {
...
+            unsafe {
+                libc::close(*pair.1);
```
danlrobertson: Same as above. It was lost in the update/re-addition.

danlrobertson:
> @danlrobertson sorry for dropping the ball on this :-( When I came back from vacation there was so much going on, and it kinda fell through the cracks...

Np... solid review

> I think it still would be useful to post the code variant using Slab somewhere, so people could try to figure out why it performs so shockingly bad...

Will do... I'll get around to adding a branch soon.

> Ironically, this PR got broken by your own PR adding FreeBSD support :-) Shouldn't be hard to fix though...

Rebased after #102 and #105 

>(And while at it, can you please update the title, to reflect that this is about mio now?...)

Updated.

Also I'd just like to point out the latest update is performing just about 100% equal with using epoll directly. Not sure if that is due to updates to mio, updates to the code here, or other updates to this repo. Either way thanks for all the input on this PR! It's been super helpful!

danlrobertson:
Sorry, accidentally hit "Close and Comment" 😄

src/platform/linux/mod.rs
antrik: (Actually, because we don't iterate `pollfds` any more...)

src/platform/linux/mod.rs
```diff
@@ -386,76 +390,84 @@ impl OsIpcChannel {
...
+                            selection_results.push(OsIpcSelectionResult::ChannelClosed(
+                                        *fd as i64))
+                        }
+                        Err(err) => return Err(err),
                     }
-                    Err(err) => return Err(err),
-                }
-                pollfd.revents = pollfd.revents & !POLLIN
+                },
+                _ => {},
```
danlrobertson: We could check `is_error` and `is_hup` similar to a previous version using epoll directly.

benches/bench.rs
```diff
@@ -170,3 +170,106 @@ fn size_22_4m(b: &mut test::Bencher) {
 fn size_23_8m(b: &mut test::Bencher) {
     bench_size(b, 8 * 1024 * 1024);
 }
+
+#[cfg(test)]
```
danlrobertson: Indeed.

benches/bench.rs
```diff
@@ -170,3 +170,106 @@ fn size_22_4m(b: &mut test::Bencher) {
...
+        gen_select_test(b, 3, 10);
+    }
+
+    #[bench]
+    fn bench_twenty_percent_5(b: &mut test::Bencher) -> () {
+        gen_select_test(b, 1, 5);
+    }
+
+    #[bench]
+    fn bench_create_and_destroy_10(b: &mut test::Bencher) -> () {
```
danlrobertson: Good call... It might be difficult to circumvent compile time optimizations since the empty set will more or less be an unused variable.

benches/bench.rs
```diff
@@ -170,3 +170,106 @@ fn size_22_4m(b: &mut test::Bencher) {
...
+        b.iter(|| {
+            let mut rx_set = IpcReceiverSet::new().unwrap();
+            add_n_rxs(&mut rx_set, 10);
+        });
+    }
+
+    #[bench]
+    fn bench_create_and_destroy_5(b: &mut test::Bencher) -> () {
```
danlrobertson: There was another line in there and a typo. Intended to be the same as above with a smaller number.

benches/bench.rs
```diff
@@ -170,3 +170,106 @@ fn size_22_4m(b: &mut test::Bencher) {
...
+        b.iter(|| {
+            let mut rx_set = IpcReceiverSet::new().unwrap();
+            add_n_rxs(&mut rx_set, 3);
+            rx_set.select().unwrap();
+        });
+    }
+
+    #[bench]
+    fn bench_manipulation(b: &mut test::Bencher) -> () {
```
danlrobertson: Yes, this test is supposed to be bench marking adding and removing receivers.

src/platform/linux/mod.rs
```diff
@@ -386,76 +390,84 @@ impl OsIpcChannel {
...
 }
 
 impl OsIpcReceiverSet {
     pub fn new() -> Result<OsIpcReceiverSet,UnixError> {
+        let fnv = BuildHasherDefault::<FnvHasher>::default();
         Ok(OsIpcReceiverSet {
-            pollfds: Vec::new(),
+            poll: Poll::new().expect("Could not create a new receiver"),
+            pollfds: HashMap::with_hasher(fnv),
+            events: Events::with_capacity(1024)
```
danlrobertson: This is what was used in all of the `mio` test cases, so that is why I used it here.

benches/bench.rs
antrik: There is `test::black_box()` should you need it -- though from my experience so far, the optimiser probably won't see through such non-trivial situations anyway...

src/platform/linux/mod.rs
antrik: Yeah, it's also in the main example. Still, since this is a pretty big number that feels like it might actually cause some noticeable overhead, I think it's worth checking whether it *needs* to be that big... (And whether it affects benchmark results.)

src/platform/unix/mod.rs
```diff
@@ -435,61 +454,57 @@ impl OsIpcReceiverSet {
...
+                            selection_results.push(OsIpcSelectionResult::ChannelClosed(fd.id))
+                        }
+                        Err(err) => return Err(err),
                     }
-                    Err(err) => return Err(err),
-                }
-                pollfd.revents = pollfd.revents & !POLLIN
+                },
+                (true, None) => panic!("Readable event for unknown token: {:?}", evt_token),
+                (false, _) => unreachable!()
```
danlrobertson: I need to do a bit more research about this across all unix OSes, but I *think* since we register all of the fd's we're watching as `readable`

src/platform/unix/mod.rs
```diff
@@ -435,61 +454,57 @@ impl OsIpcReceiverSet {
...
+                            }
+                            selection_results.push(OsIpcSelectionResult::ChannelClosed(fd.id))
+                        }
+                        Err(err) => return Err(err),
                     }
-                    Err(err) => return Err(err),
-                }
-                pollfd.revents = pollfd.revents & !POLLIN
+                },
+                (true, None) => panic!("Readable event for unknown token: {:?}", evt_token),
```
danlrobertson: Not sure how we could hit this, but as @antrik mentioned we should definitely panic if we do.

src/platform/unix/mod.rs
```diff
@@ -79,6 +81,21 @@ pub fn channel() -> Result<(OsIpcSender, OsIpcReceiver),UnixError> {
     }
 }
 
+#[derive(Clone, Copy)]
+struct FDHandle {
+    pub id: u64,
+    pub descriptor: c_int
+}
```
danlrobertson: The `OsIpcReceiverSet::pollfds` and `OsIpcReceiverSet::fdids` `HashMaps` have a 1:1 relationship. The `FDHandle` structure contains the information that would previously be contained in the two.

benches/bench.rs
```diff
@@ -170,3 +170,118 @@ fn size_22_4m(b: &mut test::Bencher) {
...
+                {
+                    let (_, rx) = ipc::channel::<()>().unwrap();
+                    rx_set.add(rx).unwrap();
+                }
+                // On select Receivers with a "ClosedChannel" event
+                // will be closed
+                rx_set.select().unwrap();
+                let (tx, rx) = ipc::channel::<()>().unwrap();
+                rx_set.add(rx).unwrap();
+                tx.send(()).unwrap();
```
antrik: I'm still confused: doesn't adding another channel and doing a send just blur the results of the channel close handling, which is what we actually want to measure here; while the other operations should already be covered by the other tests?... Or is it somehow needed for semantic reasons? Am I missing something here?...

src/platform/unix/mod.rs
```diff
@@ -415,18 +432,20 @@ impl OsIpcChannel {
...
 impl Drop for OsIpcReceiverSet {
     fn drop(&mut self) {
-        unsafe {
-            for pollfd in self.pollfds.iter() {
-                let result = libc::close(pollfd.fd);
-                assert!(thread::panicking() || result == 0);
+        let mut result = 0;
+        for (_, &FDHandle{ id: _, descriptor: fd }) in &self.pollfds {
+            unsafe {
+                result |= libc::close(fd);
```
antrik: While moving the `unsafe` block into the loop is probably a good thing (even if not strictly related to the actual change I think?), moving the `assert()` out of the loop seems like an unnecessary complication to me... Though admittedly that's rather nit-picky -- I guess that *might* just be a matter of taste?

danlrobertson:
> I'd go for something along these lines: 1/1, 1/5, 2/5, 5/5, 1/20, 5/20, 20/20, 1/100, 5/100, 20/100, 100/100.

On my list of things to do, but I'll get to it (same with the Slab example). I'm fine with adjusting the ranges.

Cargo.toml
```diff
@@ -19,5 +19,9 @@ serde = "0.8"
 uuid = {version = "0.3", features = ["v4"]}
 fnv = "1.0.3"
 
+[target.'cfg(target_os = "linux")'.dependencies]
+mio = "0.6.1"
```
danlrobertson: Good point. Hadn't updated this since that landed.

Cargo.toml
```diff
@@ -19,5 +19,9 @@ serde = "0.8"
...
+fnv = "1.0.3"
```
danlrobertson: I think it is only used in the `unix` module so perhaps it should be moved here?

src/platform/unix/mod.rs
```diff
@@ -415,80 +417,98 @@ impl OsIpcChannel {
...
 impl OsIpcReceiverSet {
     pub fn new() -> Result<OsIpcReceiverSet,UnixError> {
-        let fnv = BuildHasherDefault::<FnvHasher>::default();
+        let fnv1 = BuildHasherDefault::<FnvHasher>::default();
+        let fnv2 = BuildHasherDefault::<FnvHasher>::default();
         Ok(OsIpcReceiverSet {
             incrementor: Incrementor::new(),
-            pollfds: Vec::new(),
-            fdids: HashMap::with_hasher(fnv)
+            poll: Poll::new().expect("Could not create a new receiver"),
```
danlrobertson: AFAIK on linux this will be an abstraction of epoll_create1, so the possible source of errors will be the following

> EMFILE
The per-user limit on the number of epoll instances imposed by /proc/sys/fs/epoll/max_user_instances was encountered. See epoll(7) for further details.
ENFILE
The system limit on the total number of open files has been reached.
ENOMEM
There was insufficient memory to create the kernel object.

src/platform/unix/mod.rs
```diff
@@ -435,61 +454,57 @@ impl OsIpcReceiverSet {
...
-        self.pollfds.push(pollfd {
-            fd: fd,
-            events: POLLIN,
-            revents: 0,
-        });
-        self.fdids.insert(fd, last_index);
+        let io = EventedFd(&fd);
+        let fd_token = Token(fd as usize);
+        self.poll.register(&io, fd_token, Ready::readable(), PollOpt::level())
+            .expect(&format!("Could not register: {:?}", fd));
```
danlrobertson: Same as above, it is possible system limits will cause an error here.

Cargo.toml
antrik: Hm, OK... I don't have experience with that, but it certainly does sound reasonable :-)

That calls for a separate PR though I'd say, as it's not really related to the changes here?

src/platform/unix/mod.rs
```diff
@@ -435,61 +454,57 @@ impl OsIpcReceiverSet {
...
-                                shared_memory_regions));
-                    }
-                    Err(err) if err.channel_is_closed() => {
-                        let id = self.fdids.remove(&pollfd.fd).unwrap();
-                        unsafe {
-                            libc::close(pollfd.fd);
+        for evt in self.events.iter() {
+            let evt_token = evt.token();
+            match (evt.kind().is_readable(), self.pollfds.get(&evt_token)) {
+                (true, Some(&fd)) => {
```
danlrobertson: Makes sense

benches/bench.rs
danlrobertson: Let me look into this case again. I wrote this some time ago

src/platform/unix/mod.rs
danlrobertson: After thinking about this, I think it is more informative if it is named something like `PollEntry` over `FdEntry`.

benches/bench.rs
```diff
@@ -170,3 +170,106 @@ fn size_22_4m(b: &mut test::Bencher) {
...
+mod receiver_set {
+    use ipc_channel::ipc::{self, IpcReceiverSet};
+    use test;
+
+    fn gen_select_test(b: &mut test::Bencher, to_send: usize, n: usize) -> () {
+        let mut active = Vec::with_capacity(to_send);
+        let mut dormant = Vec::with_capacity(n - to_send);
+        let mut rx_set = IpcReceiverSet::new().unwrap();
+        let mut i = 0;
+        while i < to_send {
```
danlrobertson: Updated in the latest commit changes.

src/platform/unix/mod.rs
danlrobertson: Updated

src/platform/unix/mod.rs
danlrobertson: Added `Into<UnixError>` for `Error` for these cases.

benches/bench.rs
danlrobertson: Removed from benchmark.

src/platform/unix/mod.rs
```diff
@@ -79,6 +81,21 @@ pub fn channel() -> Result<(OsIpcSender, OsIpcReceiver),UnixError> {
...
+struct FDHandle {
+    pub id: u64,
+    pub descriptor: c_int
+}
+
+impl FDHandle {
+    pub fn new(id: u64, fd: c_int) -> FDHandle {
+        FDHandle {
+            id: id,
+            descriptor: fd
```
danlrobertson: Updated

danlrobertson:
I did some more research on using `Slab` for `OsIpcReceiverSet::pollfds`, and this iteration ([here](https://github.com/danlrobertson/ipc-channel/tree/switch_to_epoll_slab)) is definitely better than the first, but I still see no real difference in performance.

@carllerche I'd be interested if you had any input on this. Thoughts?

Here are the results from the latest benchmark tests (after a manual merge of #122)

### Use epoll directly - [Show me the code!](https://github.com/danlrobertson/ipc-channel/tree/use_raw_epoll)
```
$ cargo bench receiver_set
    Finished release [optimized] target(s) in 0.0 secs
      Running target/release/bench-cfdafcd9e0bfd194

    running 14 tests
    test receiver_set::add_and_remove_closed_receivers ... bench:       6,761 ns/iter (+/- 525)
    test receiver_set::create_and_destroy_empty_set    ... bench:         600 ns/iter (+/- 40)
    test receiver_set::create_and_destroy_set_of_10    ... bench:      26,269 ns/iter (+/- 10,149)
    test receiver_set::create_and_destroy_set_of_5     ... bench:      13,006 ns/iter (+/- 7,451)
    test receiver_set::send_on_100_of_100              ... bench:     121,379 ns/iter (+/- 4,517)
    test receiver_set::send_on_1_of_1                  ... bench:       1,137 ns/iter (+/- 46)
    test receiver_set::send_on_1_of_100                ... bench:       1,162 ns/iter (+/- 91)
    test receiver_set::send_on_1_of_20                 ... bench:       1,142 ns/iter (+/- 80)
    test receiver_set::send_on_1_of_5                  ... bench:       1,145 ns/iter (+/- 57)
    test receiver_set::send_on_20_of_100               ... bench:      22,828 ns/iter (+/- 1,822)
    test receiver_set::send_on_20_of_20                ... bench:      22,192 ns/iter (+/- 413)
    test receiver_set::send_on_2_of_5                  ... bench:       2,075 ns/iter (+/- 66)
    test receiver_set::send_on_5_of_20                 ... bench:       4,908 ns/iter (+/- 247)
    test receiver_set::send_on_5_of_5                  ... bench:       4,876 ns/iter (+/- 205)

    test result: ok. 0 passed; 0 failed; 0 ignored; 14 measured

    Running target/release/deps/ipc_channel-0d0856417f2d58dc

    running 1 test
    test platform::test::receiver_set ... ignored

    test result: ok. 0 passed; 0 failed; 1 ignored; 0 measured
```
### mio (without slab) - [Show me the code!](https://github.com/danlrobertson/ipc-channel/tree/switch_to_epoll_slab)
```
$ cargo bench receiver_set
    Finished release [optimized] target(s) in 0.0 secs
      Running target/release/bench-baa74a37c163edab

    running 14 tests
    test receiver_set::add_and_remove_closed_receivers ... bench:       9,062 ns/iter (+/- 433)
    test receiver_set::create_and_destroy_empty_set    ... bench:       2,658 ns/iter (+/- 132)
    test receiver_set::create_and_destroy_set_of_10    ... bench:      30,838 ns/iter (+/- 1,430)
    test receiver_set::create_and_destroy_set_of_5     ... bench:      16,824 ns/iter (+/- 954)
    test receiver_set::send_on_100_of_100              ... bench:     208,492 ns/iter (+/- 31,320)
    test receiver_set::send_on_1_of_1                  ... bench:       1,148 ns/iter (+/- 228)
    test receiver_set::send_on_1_of_100                ... bench:       1,145 ns/iter (+/- 205)
    test receiver_set::send_on_1_of_20                 ... bench:       1,145 ns/iter (+/- 197)
    test receiver_set::send_on_1_of_5                  ... bench:       1,146 ns/iter (+/- 200)
    test receiver_set::send_on_20_of_100               ... bench:      27,768 ns/iter (+/- 4,125)
    test receiver_set::send_on_20_of_20                ... bench:      22,190 ns/iter (+/- 3,353)
    test receiver_set::send_on_2_of_5                  ... bench:       2,087 ns/iter (+/- 362)
    test receiver_set::send_on_5_of_20                 ... bench:       5,038 ns/iter (+/- 877)
    test receiver_set::send_on_5_of_5                  ... bench:       5,038 ns/iter (+/- 843)

    test result: ok. 0 passed; 0 failed; 0 ignored; 14 measured

    Running target/release/deps/ipc_channel-30b83ba535b489f5

    running 1 test
    test platform::test::receiver_set ... ignored

    test result: ok. 0 passed; 0 failed; 1 ignored; 0 measured
```

### mio (with slab) - [Show me the code!](https://github.com/danlrobertson/ipc-channel/tree/switch_to_epoll_bench)
```
$ cargo bench receiver_set
    Finished release [optimized] target(s) in 0.0 secs
      Running target/release/bench-baa74a37c163edab

    running 14 tests
    test receiver_set::add_and_remove_closed_receivers ... bench:       9,987 ns/iter (+/- 612)
    test receiver_set::create_and_destroy_empty_set    ... bench:       3,287 ns/iter (+/- 741)
    test receiver_set::create_and_destroy_set_of_10    ... bench:      32,144 ns/iter (+/- 2,543)
    test receiver_set::create_and_destroy_set_of_5     ... bench:      17,417 ns/iter (+/- 1,028)
    test receiver_set::send_on_100_of_100              ... bench:     207,491 ns/iter (+/- 12,173)
    test receiver_set::send_on_1_of_1                  ... bench:       1,137 ns/iter (+/- 96)
    test receiver_set::send_on_1_of_100                ... bench:       1,127 ns/iter (+/- 144)
    test receiver_set::send_on_1_of_20                 ... bench:       1,147 ns/iter (+/- 70)
    test receiver_set::send_on_1_of_5                  ... bench:       1,144 ns/iter (+/- 48)
    test receiver_set::send_on_20_of_100               ... bench:      27,567 ns/iter (+/- 2,360)
    test receiver_set::send_on_20_of_20                ... bench:      21,859 ns/iter (+/- 429)
    test receiver_set::send_on_2_of_5                  ... bench:       2,084 ns/iter (+/- 111)
    test receiver_set::send_on_5_of_20                 ... bench:       4,944 ns/iter (+/- 172)
    test receiver_set::send_on_5_of_5                  ... bench:       4,931 ns/iter (+/- 233)

    test result: ok. 0 passed; 0 failed; 0 ignored; 14 measured

    Running target/release/deps/ipc_channel-30b83ba535b489f5

    running 1 test
    test platform::test::receiver_set ... ignored

    test result: ok. 0 passed; 0 failed; 1 ignored; 0 measured
```

**NB:** Do not compare these against the previously posted benchmarks. These were run on my server vs. my very old laptop.

antrik:
Quite curious that the 20 of 100 test -- just like the 100 of 100 one -- shows a considerable slowdown with `mio`; while the 20 of 20 one doesn't... (Unfortunately, the 5 of 100 case seems to be missing? I suspect it might also show a bit of a slowdown; though hopefully not too much...)

None of the other cases have a noticeable slowdown though -- so I'd say we can live with this :-)

danlrobertson:
@antrik good point. My wild guess, after a rather brief amount of time with callgrind (which showed a good chunk of time was spent in `Selector::select`), [this](https://github.com/carllerche/mio/blob/master/src/sys/unix/epoll.rs#L89-L94) looks like a reasonable culprit. If there are a reasonable number of events (e.g. 100 of 100), an O(n) situation like this is likely going to cause a slowdown. There might be other culprits, but this could be one?

Again, feedback from @alexcrichton or @carllerche on this would be really appreciated! Perhaps some of this could be moved to `mio`. I'd be happy to open an issue for `mio` and/or write a benchmark for `mio` for this case.

alexcrichton:
I did some profiling and found that [changing this to 10](https://github.com/danlrobertson/ipc-channel/blob/6dcf8d02b606d2c91b122cb63cf9ed847e214b1d/src/platform/unix/mod.rs#L449) to match the raw epoll variant made the timings much closer. I had a very difficult time seeing mio in the profile at all, it was quite noisy with allocations all over the place

src/platform/unix/mod.rs
antrik: I'm not sure how moving the assert out of the loop corresponds to the quoted passage?...

I realised though that going through the entire loop first might indeed be beneficial: this way if closing one descriptor fails, we don't leave the others leaking before triggering the thread panic... However, since this is entirely unrelated to the `mio` change as far as I can tell, I'd say this should better go into a separate PR?

Anyway, looking at the newest variant of this code, I'm beginning to wonder whether we are somehow talking past each other here... You now don't try to close at all during panic? That's definitely not right.

(Also, even though Rust asserts are not optional like in C, I still don't think it's a good idea for them to have side effects -- both for readability of the code, and of the panic messages...)

I really believe it would be best if you just replace the `Vec` with `HashMap`, and otherwise leave the code unchanged here... As a general rule of thumb, I think functional changes should be kept minimal; while cleanups and other unrelated improvements better go into separate commits -- or even separate PRs, if the changes in question have a major effect, or might be controversial in other ways...

danlrobertson:
Thanks for the comments I'll update the PR tomorrow

danlrobertson:
The lates commit changes
  - Update `drop` per comments
  - Update events size per suggestion
  - add 5/100 case in receiver_set benchmarks

@alexcrichton thanks! I found similar results. Great catch!

@antrik thanks for all the comments. The latest commits addresses the final comments. Let me know if you have any more or if I missed anything.

src/platform/unix/mod.rs
```diff
@@ -415,16 +423,16 @@ impl OsIpcChannel {
...
 
 impl Drop for OsIpcReceiverSet {
     fn drop(&mut self) {
-        unsafe {
-            for pollfd in self.pollfds.iter() {
-                let result = libc::close(pollfd.fd);
-                assert!(thread::panicking() || result == 0);
+        for &PollEntry { id: _, fd } in self.pollfds.values() {
+            unsafe {
+                assert!(thread::panicking() || libc::close(fd) == 0);
```
antrik: I'm sorry to say that this is still not right :-( Again, this code won't even try to close the FDs during a panic, which is wrong. (And also, putting actual functionality in an `assert!()` rather than just checks is not a good idea.)

Really, there are good reasons why this code is arranged this way...

src/platform/unix/mod.rs
```diff
@@ -435,61 +443,76 @@ impl OsIpcReceiverSet {
         let fnv = BuildHasherDefault::<FnvHasher>::default();
         Ok(OsIpcReceiverSet {
             incrementor: Incrementor::new(),
-            pollfds: Vec::new(),
-            fdids: HashMap::with_hasher(fnv)
+            poll: match Poll::new() {
+                Ok(poll) => poll,
+                Err(e) => { return Err(e.into()); }
```
antrik: I somehow missed this before (sorry): but if you implemented the `From` trait instead of the `Into` one -- like the other back-ends do -- you could just use `try!()` here I believe?...

src/platform/unix/mod.rs
```diff
@@ -435,61 +443,76 @@ impl OsIpcReceiverSet {
...
+        match self.poll.register(&io,
+                                 fd_token,
+                                 Ready::readable(),
+                                 PollOpt::level()) {
+            Ok(_) => {
+                self.pollfds.insert(fd_token, poll_entry);
+                Ok(last_index)
+            },
+            Err(e) => {
+                Err(e.into())
```
antrik: This could also use `try!()` I think? Though the win here is slightly less clear than above I guess...

danlrobertson:
@antrik Ah, true. Changed the `Into<io::Error>` impl to `From<io::Error>` and used `try!`

antrik:
@danlrobertson I'm beginning to wonder whether you perhaps missed my other comment, about `close()` handling still not being right?...

(I posted it separately from the other remarks by accident.)

danlrobertson:
Updated to not be updated :smile: (reverted the drop implementation to be more like the original)

antrik:
Hrm, apparently we have intermittent failures in the "router" mechanism? :-(

That is unrelated to this PR though. I think it looks good now :-)

danlrobertson:
Force pushed to trigger a build, and there seem to be build machine errors. I'll try another force push later.

pcwalton:
@bors-servo: r+

Awesome simplification :)

notriddle:
Failed to update the crates.io registry.

@bors-servo retry

notriddle:
One more time: @bors-servo retry

antrik:
I guess another retry is in order? Or is this CI issue more permanent?...

notriddle:
```
Build started
Environment variable name or value is too long.
```

danlrobertson:
\o/ Thanks for all the comments and critiques in the reviews.

emilio:
Thank _you_ for doing this, and for the patience to get this landed!

nox:
Such a big change should have been tested on Servo.

https://github.com/servo/servo/pull/15537#issuecomment-279649494

PR #95, opened 2016-08-09
Added unit test for reentrancy.

asajeffrey:
IPC channels are reentrant, in that if we call `c.send(a)` and serializing `a` calls `c.send(b)`, then the channel contents are `b` then `a`. (This comes up in the case of logging over channels, since serialization may include logging.)

This PR adds a unit test to make sure we don't break reentrancy.

Discussion with @nox on IRC.

### DIFF
--- a/src/test.rs
+++ b/src/test.rs
@@ -11,6 +11,8 @@ use ipc::{self, IpcReceiver, IpcReceiverSet, IpcSender, IpcSharedMemory};
 use ipc::{OpaqueIpcSender};
 use router::ROUTER;
 use libc;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use std::cell::RefCell;
 use std::iter;
 use std::ptr;
 use std::sync::Arc;
@@ -19,6 +21,7 @@ use std::thread;
 
 #[cfg(not(any(feature = "force-inprocess", target_os = "windows", target_os = "android")))]
 use ipc::IpcOneShotServer;
+
 #[cfg(not(any(feature = "force-inprocess", target_os = "windows", target_os = "android")))]
 use std::io::Error;
 
@@ -385,4 +388,41 @@ fn test_so_linger() {
         Err(e) => { panic!("err: `{}`", e); }
     };
     assert_eq!(val, 42);
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+struct HasWeirdSerializer (Option<String>);
+
+thread_local! { static WEIRD_CHANNEL: RefCell<Option<IpcSender<HasWeirdSerializer>>> = RefCell::new(None) }
+
+impl Serialize for HasWeirdSerializer {
+    fn serialize<S>(&self, serializer: &mut S) -> Result<(), S::Error>
+        where S: Serializer
+    {
+        if self.0.is_some() {
+            WEIRD_CHANNEL.with(|chan| { chan.borrow().as_ref().unwrap().send(HasWeirdSerializer(None)).unwrap(); });
+        }
+        self.0.serialize(serializer)
+    }
+}
+
+impl Deserialize for HasWeirdSerializer {
+    fn deserialize<D>(deserializer: &mut D) -> Result<Self, D::Error>
+        where D: Deserializer
+    {
+        Ok(HasWeirdSerializer(try!(Deserialize::deserialize(deserializer))))
+    }
+}
+
+#[test]
+fn test_reentrant() {
+    let null = HasWeirdSerializer(None);
+    let hello = HasWeirdSerializer(Some(String::from("hello")));
+    let (sender, receiver) = ipc::channel().unwrap();
+    WEIRD_CHANNEL.with(|chan| { *chan.borrow_mut() = Some(sender.clone()); });
+    sender.send(hello.clone()).unwrap();
+    assert_eq!(null, receiver.recv().unwrap());
+    assert_eq!(hello, receiver.recv().unwrap());
+    sender.send(null.clone()).unwrap();
+    assert_eq!(null, receiver.recv().unwrap());
 }


asajeffrey:
cc @pcwalton

src/test.rs
```diff
@@ -49,6 +50,30 @@ type PersonAndOpaqueSender = (Person, OpaqueIpcSender);
 type PersonAndReceiver = (Person, IpcReceiver<Person>);
 type PersonAndSharedMemory = (Person, IpcSharedMemory);
 
+#[derive(Clone, Debug, Eq, PartialEq)]
+struct HasWeirdSerializer (Option<String>);
+
+lazy_static! { static ref WEIRD_CHANNEL: (IpcSender<HasWeirdSerializer>, IpcReceiver<HasWeirdSerializer>) = ipc::channel().unwrap(); }
```
antrik: A non-reentrant testcase for reentrance? No that's just lazy! ;-)

I guess it's not really a problem, unless the test harness ever grows an ability to run the same test multiple times in parallel, which seems rather hypothetical...

antrik:
For what it's worth, this looks good to me.

(I'm happy to see a testcase being added for obscure behaviour someone will be relying on -- a very good practice that probably should be employed more often :-) )

jdm:
Travis is displeased:
```
     Running `rustc src/lib.rs --crate-name ipc_channel -g --test -C metadata=f57b0507e715624e -C extra-filename=-f57b0507e715624e --out-dir /Users/travis/build/servo/ipc-channel/target/debug/deps --emit=dep-info,link -L dependency=/Users/travis/build/servo/ipc-channel/target/debug/deps --extern rand=/Users/travis/build/servo/ipc-channel/target/debug/deps/librand-49a08859d086fffe.rlib --extern lazy_static=/Users/travis/build/servo/ipc-channel/target/debug/deps/liblazy_static-359f5533c970cd71.rlib --extern crossbeam=/Users/travis/build/servo/ipc-channel/target/debug/deps/libcrossbeam-95e5cb210400af41.rlib --extern bincode=/Users/travis/build/servo/ipc-channel/target/debug/deps/libbincode-6bdca1673f724e3d.rlib --extern libc=/Users/travis/build/servo/ipc-channel/target/debug/deps/liblibc-1bd8847afb79f283.rlib --extern serde=/Users/travis/build/servo/ipc-channel/target/debug/deps/libserde-652fd324488db42d.rlib`

<lazy_static macros>:2:32: 3:65 error: the trait bound `std::cell::Cell<u32>: std::marker::Sync` is not satisfied [E0277]

<lazy_static macros>:2 use std :: sync :: ONCE_INIT ; static mut $ NAME : $ crate :: lazy :: Lazy < $

                                                      ^

<lazy_static macros>:21:1: 21:40 note: in this expansion of __lazy_static_create! (defined in <lazy_static macros>)

<lazy_static macros>:4:1: 5:75 note: in this expansion of lazy_static! (defined in <lazy_static macros>)

src/test.rs:56:1: 56:135 note: in this expansion of lazy_static! (defined in <lazy_static macros>)

<lazy_static macros>:2:32: 3:65 help: run `rustc --explain E0277` to see a detailed explanation

<lazy_static macros>:2:32: 3:65 note: `std::cell::Cell<u32>` cannot be shared between threads safely 

<lazy_static macros>:2:32: 3:65 note: required because it appears within the type `platform::os::OsIpcReceiver` 

<lazy_static macros>:2:32: 3:65 note: required because it appears within the type `ipc::IpcReceiver<test::HasWeirdSerializer>` 

<lazy_static macros>:2:32: 3:65 note: required because it appears within the type `(ipc::IpcSender<test::HasWeirdSerializer>, ipc::IpcReceiver<test::HasWeirdSerializer>)` 

<lazy_static macros>:2:32: 3:65 note: required by `lazy_static::lazy::Lazy` 

error: aborting due to previous error 

error: Could not compile `ipc-channel`.
```

asajeffrey:
Not sure why I'm not getting this error on my end... I'll put it in a rw mutex or some such and retry.

antrik:
@asajeffrey on what platform(s) have you tested this? The definition of `IpcReceiver` is presently kinda wrong on the `linux` and `inprocess` backends (only the `macos` one gets it right): the definitions formally look thread-safe, which is not really true -- see discussion on https://github.com/servo/ipc-channel/pull/89 . I have some work in progress to fix this, which I haven't quite wrapped up yet; but once that's in, your code would fail equally on all platforms.

asajeffrey:
The test isn't to see if it's thread-safe, it's to see if it's reentrant. Since IPC channels serialize to a fresh buffer each time, it should be reentrant.

antrik:
@asajeffrey well, that's not what the test is about -- but since `lazy_static` creates a global variable, Rust enforces that whatever it holds must be thread-safe. That's not really the case for `IpcReceiver`.

asajeffrey:
Indeed, since it's a global it has to be sync. Rather ironically this is exactly the use case for reentrant locks, which is what got is here in the first place. Probably easiest just to make it thread-local. I'm not at my desk today, I'll do this tomorrow.

asajeffrey:
I made Travis happy by making the global thread-local.

src/test.rs
```diff
@@ -49,6 +51,30 @@ type PersonAndOpaqueSender = (Person, OpaqueIpcSender);
...
+}
+
+impl Deserialize for HasWeirdSerializer {
+    fn deserialize<D>(deserializer: &mut D) -> Result<Self, D::Error>
+        where D: Deserializer
+    {
+        Ok(HasWeirdSerializer(try!(Deserialize::deserialize(deserializer))))
+    }
+}
+
```
antrik: Personally I think it would be more readable to put all this stuff right before the (only) test case using it...

src/test.rs
```diff
@@ -384,3 +410,17 @@ fn test_so_linger() {
...
     assert_eq!(val, 42);
 }
+
+#[test]
+fn test_reentrant() {
+    let null = HasWeirdSerializer(None);
+    let hello = HasWeirdSerializer(Some(String::from("hello")));
+    let (sender, receiver) = ipc::channel().unwrap();
+    let sender1 = sender.clone();
+    WEIRD_CHANNEL.with(|chan| { *chan.borrow_mut() = Some(sender1); });
```
antrik: Nitpick: since this is not a move closure, it seem circuitous to do the `clone()` beforehand -- I think it would be cleaner simply to do `Some(sender.clone())` within the closure...

This is mostly just a general hint though -- as we are talking about test code here, I wouldn't insist on fixing such a minor stylistic wrinkle :-)

src/test.rs
asajeffrey: Fair enough.

src/test.rs
asajeffrey: Okay, I'll make this change.

antrik:
@asajeffrey thanks :-) Please squash the commits -- and then we just need someone with official reviewer status to approve it...

asajeffrey:
Squashed.

antrik:
Meh, there is an intermittent failure it seems? This definitely shouldn't be affected by these changes...

asajeffrey:
Yes, that is odd... timeouts on unrelated tests?

antrik:
Just to be clear: I really don't think this test failure has anything to do with this PR. Either the test case in question has intermittent failures, or there was some one-time hiccup on Travis.

@pcwalton r?

pcwalton:
Looks good once the conflicts are resolved.

PR #99, opened 2016-08-27
Remove mach_msg retry loop.

jrmuizel:
The idiomatic solution to dealing with the trailer bytes
is to just add the maximum trailer size to msg size returned by the
kernel.


pcwalton:
Nice!

@bors-servo: r+

PR #102, opened 2016-09-18
Support Freebsd

danlrobertson:
Add support for FreeBSD and rename `linux` module to `unix`.

**NB:** Some tests are still currently failing on freebsd. I believe most of which have to do with system limits e.g. (`platform::medium_data` blocks on a `send_followup_fragment` due to system limits and never recovers).


src/platform/unix/mod.rs
```diff
@@ -953,21 +976,32 @@ fn is_socket(fd: c_int) -> bool {
...
+#[cfg(target_os = "linux")]
+unsafe fn CMSG_DATA(cmsg: *mut cmsghdr) -> *mut c_void {
+    cmsg.offset(1) as *mut c_void
 }
 
 #[allow(non_snake_case)]
-fn CMSG_ALIGN(length: size_t) -> size_t {
-    (length + mem::size_of::<size_t>() - 1) & !(mem::size_of::<size_t>() - 1)
+#[cfg(target_os = "freebsd")]
+unsafe fn CMSG_DATA(cmsg: *mut cmsghdr) -> *mut c_void {
```
danlrobertson: Ensure proper alignment. `cmsg.offset(1)` on freebsd is the equivalent of `addq 0xc, %rcx`, but it should be `addq 0x10, %rcx`.

src/platform/unix/mod.rs
```diff
@@ -192,13 +238,13 @@ impl OsIpcSender {
...
-                    let cmsg_buffer = libc::malloc(CMSG_SPACE(cmsg_length)) as *mut cmsghdr;
-                    (*cmsg_buffer).cmsg_len = CMSG_LEN(cmsg_length);
+                    let cmsg_buffer = libc::malloc(CMSG_SPACE(cmsg_length) as usize) as *mut cmsghdr;
+                    (*cmsg_buffer).cmsg_len = CMSG_LEN(cmsg_length as SockLen);
                     (*cmsg_buffer).cmsg_level = libc::SOL_SOCKET;
                     (*cmsg_buffer).cmsg_type = SCM_RIGHTS;
 
                     ptr::copy_nonoverlapping(fds.as_ptr(),
-                                             cmsg_buffer.offset(1) as *mut c_int,
+                                             CMSG_DATA(cmsg_buffer) as *mut c_int,
```
danlrobertson: Use `CMSG_DATA` instead of `ptr::offset`

antrik:
@danlrobertson I'm not sure if/when I get to look at the changes in detail; so some quick comments for now:

1. Formalities: I believe the changes you commented on here should actually be separate commits, with the comments as commit messages. The comments here won't be terribly useful when someone tries to make sense of the Git changelog at some point in the future...

2. The `CMSG_DATA`/`CMSG_ALIGN`/`CMSG_LEN` macros are direct copies of macros from Linux-specific bits of glibc, only translated to Rust -- so they obviously need system-specific variants. The code *invoking* them on the other hand should be portable I think...

3. This might be considered out of scope -- but I have been thinking for a while that these extremely system-specific details should probably not be present in the code of `ipc-channel` at all, but rather covered by external crates, such as `nix` and/or `libc` -- though quite probably this will require some upstream enhancements to said crates...

danlrobertson:
@antrik thanks for the quick reply. Don't feel rushed.

1) Shouldn't be a problem.

2) The changes I made to the FreeBSD specific `CMSG_DATA` function was also a translation of the C macro. I don't see how the current code can be written to be more portable without something along the lines of `3`.

3) I completely agree. I was surprised this wasn't in `libc`

l0kod:
About the CMSG_DATA, you may be interested in https://github.com/stemjail/fdpass-rs

danlrobertson:
@antrik: Updated to include 2 commits. 1) move static and const declarations and 2) support freebsd

@l0kod: per stemjail/fdpass-rs#1, it doesn't seem to be currently working on FreeBSD. Let me know if missed something.

antrik:
@l0kod I briefly considered `fdpass-rs` when I first read about it; but it seems to me `nix` with its much broader scope would be more appropriate for `ipc-channel`...

src/platform/unix/mod.rs
```diff
@@ -0,0 +1,1029 @@
+// Copyright 2015 The Servo Project Developers. See the COPYRIGHT
```
antrik: Something went wrong with your commit: it introduces both ` src/platform/unix/mod.rs` and `src/platform/unix/linux/mod.rs` (with identical contents) in place of the original `src/platform/linux/mod.rs`...

danlrobertson:
Thanks for the very thorough review. I'll make the changes and comment when I make the update. I'll also start looking into a better home for [`cmsg`](http://man7.org/linux/man-pages/man3/cmsg.3.html)

src/platform/unix/mod.rs
danlrobertson: I'll fix this in the update... My bad.

src/platform/unix/linux/mod.rs
```diff
@@ -953,21 +976,32 @@ fn is_socket(fd: c_int) -> bool {
...
+fn CMSG_ALIGN(length: size_t) -> SockLen {
+    ((length + mem::size_of::<size_t>() - 1) & !(mem::size_of::<size_t>() - 1)) as SockLen
+}
 
 #[allow(non_snake_case)]
-fn CMSG_LEN(length: size_t) -> size_t {
-    CMSG_ALIGN(mem::size_of::<cmsghdr>()) + length
+#[cfg(target_os = "linux")]
+unsafe fn CMSG_DATA(cmsg: *mut cmsghdr) -> *mut c_void {
+    cmsg.offset(1) as *mut c_void
```
danlrobertson: Didn't think about that. That would make this a bit more simplistic. I'd definitely be okay with making this change

src/platform/unix/linux/mod.rs
```diff
@@ -953,21 +976,32 @@ fn is_socket(fd: c_int) -> bool {
 
 // FFI stuff follows:
 
-const SCM_RIGHTS: c_int = 0x01;
+#[allow(non_snake_case)]
+fn CMSG_LEN(length: SockLen) -> SockLen {
```
danlrobertson: Agreed :+1:

src/platform/unix/linux/mod.rs
```diff
@@ -26,18 +26,64 @@ use std::thread;
...
+#[inline]
+unsafe fn new_sockaddr_un(path: *const c_char) -> (sockaddr_un, usize) {
+    let mut sockaddr = sockaddr_un {
+        sun_family: libc::AF_UNIX as u16,
+        sun_path: [ 0; 108 ],
+    };
+    libc::strncpy(sockaddr.sun_path.as_mut_ptr(),
+                  path,
+                  sockaddr.sun_path.len() - 1);
+    let len = mem::size_of::<sockaddr_un>() - 104 + (libc::strlen(sockaddr.sun_path.as_ptr()) as usize);
```
danlrobertson: Yes, this is an error. `linux` would be `- 108`, while `freebsd` would be `- 104`

src/platform/unix/linux/mod.rs
```diff
@@ -26,18 +26,64 @@ use std::thread;
...
+type IovLen = usize;
+#[cfg(target_os="linux")]
+type SockLen = usize;
+#[cfg(target_os="linux")]
+#[inline]
+unsafe fn new_sockaddr_un(path: *const c_char) -> (sockaddr_un, usize) {
+    let mut sockaddr = sockaddr_un {
+        sun_family: libc::AF_UNIX as u16,
+        sun_path: [ 0; 108 ],
+    };
```
danlrobertson: I'll give it a shot.

src/platform/unix/linux/mod.rs
```diff
@@ -26,18 +26,64 @@ use std::thread;
...
+#[cfg(not(target_os="android"))]
+const TEMP_FILE_TEMPLATE: &'static str = "/tmp/ipc-channel-shared-memory.XXXXXX";
+
+
+#[cfg(target_os="linux")]
+type IovLen = usize;
+#[cfg(target_os="linux")]
+type SockLen = usize;
+#[cfg(target_os="linux")]
+#[inline]
```
danlrobertson: :+1:

src/platform/unix/linux/mod.rs
```diff
@@ -26,18 +26,64 @@ use std::thread;
...
+const TEMP_FILE_TEMPLATE: &'static str = "/sdcard/servo/ipc-channel-shared-memory.XXXXXX";
+
+#[cfg(not(target_os="android"))]
+const TEMP_FILE_TEMPLATE: &'static str = "/tmp/ipc-channel-shared-memory.XXXXXX";
+
+
+#[cfg(target_os="linux")]
+type IovLen = usize;
+#[cfg(target_os="linux")]
+type SockLen = usize;
```
danlrobertson: That's what I was hoping. However, [socklen_t:linux = u32](https://doc.rust-lang.org/libc/x86_64-unknown-linux-gnu/libc/type.socklen_t.html) but [msghdr.msg_controllen = size_t = usize](https://doc.rust-lang.org/libc/x86_64-unknown-linux-gnu/libc/struct.msghdr.html). This works on FreeBSD though, and was what I originally did. Perhaps this is another change that should be made to `libc`

src/platform/unix/linux/mod.rs
```diff
@@ -26,18 +26,64 @@ use std::thread;
 
 const MAX_FDS_IN_CMSG: u32 = 64;
 
+const SCM_RIGHTS: c_int = 0x01;
+
+// The value Linux returns for SO_SNDBUF
+// is not the size we are actually allowed to use...
+// Empirically, we have to deduct 32 bytes from that.
+const RESERVED_SIZE: usize = 32;
```
danlrobertson: I want to say yes, but I'll double check.

src/platform/unix/linux/mod.rs
```diff
@@ -26,18 +26,64 @@ use std::thread;
 
 const MAX_FDS_IN_CMSG: u32 = 64;
 
+const SCM_RIGHTS: c_int = 0x01;
```
danlrobertson: It apears as if everything from `cmsg` is not defined in `libc`. It definitely seems like they should be defined somewhere like `libc` or `nix`.

antrik: `libc` I'd say. `libc` is for the raw FFI interfaces; while `nix` is for safe wrappers around them. Using `nix` to abstract all these dirty details would be nice to have, but it's a second (much bigger) step...

src/platform/unix/linux/mod.rs
antrik: Ah, so for some reason the POSIX committee decided to abuse `socklen_t` for something that is clearly *not* in any way related to socket address lengths; while `glibc` decided to stick with the more appropriate `size_t`... This is so retarded :-(

Calling the type `SockLen` is very confusing though, as it has no bearing on the primary use of `socklen_t` -- better call it `MsgControllen`, or something along these lines. Also, it's probably better to define it as `size_t` rather than `usize`, to match the `libc` definition.

danlrobertson:
> Personally, I'd prefer even finer-grained commits

Updated, and I must admit you were right... It makes a lot more sense this way.

> Calling the type SockLen is very confusing though.. better call it MsgControllen

Updated.

> Also, it's probably better to define it as size_t rather than usize, to match the libc definition.

Updated.

> I suspect we could make this code [new_sockaddr_un] generic, by using mem::zeroed() and then setting only the fields we need

Updated.

src/platform/mod.rs
```diff
@@ -8,8 +8,9 @@
 // except according to those terms.
 
 mod os {
-    #[cfg(all(not(feature = "force-inprocess"), target_os = "linux"))]
-    include!("linux/mod.rs");
+    #[cfg(all(not(feature = "force-inprocess"), any(target_os = "linux",
+                                                    target_os = "freebsd")))]
```
antrik: Personally, I'd add the `freebsd` case only right before or along with actually adding FreeBSD-specific code, after the generic portability improvements... But this way is fine too I guess :-)

src/platform/unix/mod.rs
```diff
@@ -39,11 +39,25 @@ const TEMP_FILE_TEMPLATE: &'static str = "/sdcard/servo/ipc-channel-shared-memor
...
+type MsgControlLen = socklen_t;
+#[cfg(target_os = "freebsd")]
+type SunFamily = u8;
+
 unsafe fn new_sockaddr_un(path: *const c_char) -> (sockaddr_un, usize) {
     let mut sockaddr: sockaddr_un = mem::zeroed();
     libc::strncpy(sockaddr.sun_path.as_mut_ptr(),
                   path, sockaddr.sun_path.len() - 1);
-    sockaddr.sun_family = libc::AF_UNIX as u16;
+    sockaddr.sun_family = libc::AF_UNIX as SunFamily;
```
antrik: Can't you just use `sa_family_t` here as I suggested?...

src/platform/unix/mod.rs
```diff
@@ -961,7 +954,7 @@ fn CMSG_LEN(length: size_t) -> size_t {
 }
 
 #[allow(non_snake_case)]
-fn CMSG_DATA(cmsg: *mut cmsghdr) -> *mut c_void {
+unsafe fn CMSG_DATA(cmsg: *mut cmsghdr) -> *mut c_void {
```
antrik: I think that belongs with the previous commit?...

src/platform/unix/mod.rs
```diff
@@ -39,6 +39,16 @@ const TEMP_FILE_TEMPLATE: &'static str = "/sdcard/servo/ipc-channel-shared-memor
 #[cfg(not(target_os="android"))]
 const TEMP_FILE_TEMPLATE: &'static str = "/tmp/ipc-channel-shared-memory.XXXXXX";
 
+unsafe fn new_sockaddr_un(path: *const c_char) -> (sockaddr_un, usize) {
+    let mut sockaddr: sockaddr_un = mem::zeroed();
+    libc::strncpy(sockaddr.sun_path.as_mut_ptr(),
+                  path, sockaddr.sun_path.len() - 1);
+    sockaddr.sun_family = libc::AF_UNIX as u16;
```
antrik: I'd split the commits a bit differently: first only split out `new_sockaddr_un()`, without any change to the actual code moved in there, so it's a pure refactoring; and only after that one or more commits that actually make it platform-agnostic, but introducing `mem::zeroed()`, `sa_family_t`, etc.

(Not insisting though -- probably it's clear enough that way as well...)

src/platform/unix/mod.rs
```diff
@@ -39,6 +39,16 @@ const TEMP_FILE_TEMPLATE: &'static str = "/sdcard/servo/ipc-channel-shared-memor
...
+    let len = mem::size_of::<c_short>() + (libc::strlen(sockaddr.sun_path.as_ptr())
+                                           as usize);
```
antrik: While this happens to work on Linux and FreeBSD, I don't think it's really generic... As I said, this would probably require some pointer arithmetic -- but I still think the *right* way to handle this is to just use `sizeof sockaddr_un`, like in the Linux man page example, rather than trying to determine the actually used size dynamically.

src/platform/unix/mod.rs
danlrobertson: Done! in the latest batch of commits

src/platform/unix/mod.rs
danlrobertson: Good point. Used `sizeof(sockaddr_un)` in the latest batch of commits.

antrik:
Apart from the pending question regarding handling of platform conditionals, I'd say this is good to go :-)

myfreeweb:
Thank you for doing this work! Got [Webrender working on FreeBSD](https://unrelenting.technology/notes/2016-10-21-21-16-04) :)

danlrobertson:
Updated with the suggested patch splits.

antrik:
Great, let's get this landed :-)

Issue #83, opened 2016-06-29
Trailing data with `bytes_channel` on MacOS


antrik:
As evidenced by https://github.com/servo/ipc-channel/pull/79/commits/988a503cd6c9531d49861c81d35be8f4e4577d17 , when transferring raw data with a length not a multiple of 4 through a `bytes_channel`, on the MacOS backend, the data is padded with `0`s. Sooner or later this is going to bite someone.

Gankro:
@jdm suggested this can be prioritized. It seems like it's a fairly easy/mentorable bug, just requires someone to dig in and find the right place to handle this and pipe the info through.

The rust code is very explicitly doing this rounding, so we know when it's happening. That code also seems to be a bit of a mess; could use some TLC.

PR #104, opened 2016-09-20
Remove received_data.truncate() calls from platform/test.rs

vvuk:
Let's see what happens!


antrik:
It fails -- see https://github.com/servo/ipc-channel/issues/83

PR #105, opened 2016-10-01
Use autoincrementing number for handler ids

danlrobertson:
Use `Uuid` instead of `i64` for the handler keys to ensure file descriptor number reuse doesn't cause errors, and make sure to close file descriptors.

**NB:** The scope of the PR has changed to change the handler id's from a platform specific `i64` to an autoincrementing `u64`. `macos` has not yet been ported, and probably suffers from the same leak as the `unix` module did prior to this PR

### DIFF
--- a/src/platform/unix/mod.rs
+++ b/src/platform/unix/mod.rs
@@ -14,7 +14,7 @@ use libc::{SO_LINGER, S_IFMT, S_IFSOCK, c_char, c_int, c_void, getsockopt};
 use libc::{iovec, mkstemp, mode_t, msghdr, nfds_t, off_t, poll, pollfd, recvmsg, sendmsg};
 use libc::{setsockopt, size_t, sockaddr, sockaddr_un, socketpair, socklen_t, sa_family_t};
 use std::cmp;
-use std::collections::{HashMap, HashSet};
+use std::collections::HashMap;
 use std::ffi::{CStr, CString};
 use std::fmt::{self, Debug, Formatter};
 use std::hash::BuildHasherDefault;
@@ -453,7 +453,6 @@ impl OsIpcReceiverSet {
             return Err(UnixError::last())
         }
 
-        let mut hangups = HashSet::new();
         for pollfd in self.pollfds.iter_mut() {
             if (pollfd.revents & POLLIN) != 0 {
                 match recv(pollfd.fd, BlockingMode::Blocking) {
@@ -465,9 +464,11 @@ impl OsIpcReceiverSet {
                                 shared_memory_regions));
                     }
                     Err(err) if err.channel_is_closed() => {
-                        hangups.insert(pollfd.fd);
-                        selection_results.push(OsIpcSelectionResult::ChannelClosed(
-                                    *self.fdids.get(&pollfd.fd).unwrap()))
+                        let id = self.fdids.remove(&pollfd.fd).unwrap();
+                        unsafe {
+                            libc::close(pollfd.fd);
+                        }
+                        selection_results.push(OsIpcSelectionResult::ChannelClosed(id))
                     }
                     Err(err) => return Err(err),
                 }
@@ -475,9 +476,11 @@ impl OsIpcReceiverSet {
             }
         }
 
-        if !hangups.is_empty() {
-            self.pollfds.retain(|pollfd| !hangups.contains(&pollfd.fd));
-        }
+        // File descriptors not in fdids are closed channels, and the descriptor
+        // has been closed. This must be done after we have finished iterating over
+        // the pollfds vec.
+        let fdids = &self.fdids;
+        self.pollfds.retain(|pollfd| fdids.contains_key(&pollfd.fd));
 
         Ok(selection_results)
     }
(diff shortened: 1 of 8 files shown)


vvuk:
Instead of uuid, can we just have a i64 id that we just atomically increment when we need one?  Would also mean no changes to anything but the platform backends.

danlrobertson:
Valid point. We could do that, but imo if we did that we should definitely implement some sort of id reuse. The creation of the id reuse algorithm could take as much effort as switching Uuid, depending on the route taken.

vvuk:
Why do we need id reuse?  If a program allocated a new id every millisecond, it would take ~585 million years to exaust 2^64. I think that's a pretty good definition of "someone else's problem" ;)  (1 per ns gives us 585 years, which is *slightly* more concerning, but only slightly.)

danlrobertson:
Ha very vary true... I though about that, and AFAIK you'd have to exhaust ((2 ^ 32) - 1) due to the use of `i64` and the two's complement, so you'd end up with ~292 years which from what I've heard is about twice to three time as long as it would take to generate a duplicate `Uuid` (from what I've read on the internets). If we switched to `u64` for the id we'd be at ~585.

If the consensus is to use an auto incrementing `i64` I'm 100% okay with that. I'm just hyper paranoid about writing code that allows for overflow.

antrik:
It's `2^63-1` till integer overflow -- actual collision would indeed be exactly `2^64` I believe. You should be able to avoid the overflow, to squeeze out that last bit (literally), by using `u64` internally, and casting to `i64` as necessary.

(For production builds, it should produce the exact same machine code I think?)

Either way, I think it will be more efficient and less invasive than using `UUID`s.

I'd say the actual `FD` closing should probably go into a separate patch. (The ID change lays the groundwork by decoupling the handles from the `FD`s; while the main change builds on top of this to fix the leak AIUI...)

Changing the handling of `r_id` to use `Option<>` can probably also go into a separate patch before the main ID handling change, for easier review... Unless it's too much hassle to untangle it I guess.

danlrobertson:
>  using u64 internally, and casting to i64 as necessary.

Good point. I'll use this in the update

> I'd say the actual FD closing should probably go into a separate patch.

Np. Nothing `git rebase` can't handle.

Thanks @antrik and @vvuk! I'll have some time this weekend and push the update using `u64`.

danlrobertson:
Updated to use an auto-incrementing `u64`

danlrobertson:
Thanks again, great critiques! I clearly need to work on better commit messages  :smile:.

src/platform/unix/mod.rs
```diff
@@ -465,19 +464,23 @@ impl OsIpcReceiverSet {
...
                 pollfd.revents = pollfd.revents & !POLLIN
             }
         }
 
-        if !hangups.is_empty() {
-            self.pollfds.retain(|pollfd| !hangups.contains(&pollfd.fd));
-        }
+        // Avoid the use of `self` in closue to avoid errors due to the
+        // mutable borrow of `self` at line 483
+        let ids = &self.fdids;
```
KiChjang: What happens if instead of doing an inline closure in line 483, you make a let-binding of a closure and pass it in as an argument to `retain`?

danlrobertson: Good idea! but sadly no :(

antrik: I think I understand the situation; I don't understand though why you consider it a problem?...

It's not exactly uncommon to create temporary bindings to get around borrow conflicts. I'm not sure this even deserves a comment: it doesn't seem like an obscure situation that requires much of an explanation?...

(BTW, including line numbers in a comment is a bad idea: these are bound to change :-) )

On the other hand, I for my part would be tempted to add a comment explaining that the entries that got removed from `fdids`, because the channels got closed, also need to be purged from `pollfds`. (And perhaps even mention that we can only do that after we finished iterating `pollfds`?...) These are the kind of things that are probably not immediately obvious even to an experienced Rust programmer looking at this code for the first time.

(I am aware that the existing `ipc-channel` code doesn't have much in the manner of such explanations... Which I personally consider a pity.)

On a more nit-picky note, I'd drop the blank line after the temporary binding -- these two lines very much belong together. Also, I think it would actually be preferable to name the temporary as `fdids` as well...

danlrobertson: > (BTW, including line numbers in a comment is a bad idea: these are bound to change :-) )

Ha true! good call... Removed the line numbers

I also added a comment above the retain detailing the reason for the placement.

antrik:
Looks good to me :-) Now we just need someone to approve it...

src/platform/unix/mod.rs
antrik: BTW, just for explanation (in case it isn't clear): the issue here isn't the borrow checker being dumb, but rather it's about the way how closures are handled. They automatically close over any binding used within the closure's code -- in this case the binding is `self`. The compiler doesn't try to figure out that you are actually accessing only one sub-field. If you want only the sub-field to be closed over, you have to create an explicit binding for it... Which is precisely what you did :-)

danlrobertson:
Cool beans! I'll move back onto servo/ipc-channel#94

emilio:
Thanks for doing this work, it's awesome :)

PR #107, opened 2016-10-04
inprocess: Don't wrap `OsIpcReceiver` to make it `Sync`

antrik:
Wrap the inprocess receiver in a simple `RefCell<>` instead of
`Arc<Mutex<>>` (again).

Adding the mutex here is not a good idea: it only adds overhead and
potential errors, since not all of the other backends have a `Sync`
receiver either; nor does the `mpsc` mechanism `ipc-channel` is modelled
after. In fact it fundamentally violates the "single receiver" aspect of
`mpsc`, as outlined in
https://github.com/servo/ipc-channel/pull/89#issuecomment-234394535

Users can still wrap it explicitly if they need to, and know it doesn't
introduce incorrect behaviour in the specific use case.

This change mostly reverts the `Receiver` part of
30b20244dc83e61f9ba09b6e3db13a8ff7ac4665 ; it doesn't re-introduce the
explicit `Sync` declaration for `Receiver` though, as that was/is
clearly not true without the Mutex -- nor really desirable, as explained
above. This change also doesn't touch the `Sender` part, which is a
different story.

### DIFF
--- a/src/platform/inprocess/mod.rs
+++ b/src/platform/inprocess/mod.rs
@@ -60,13 +60,13 @@ pub fn channel() -> Result<(OsIpcSender, OsIpcReceiver),MpscError> {
 }
 
 pub struct OsIpcReceiver {
-    receiver: Arc<Mutex<Option<mpsc::Receiver<MpscChannelMessage>>>>,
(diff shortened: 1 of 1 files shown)


jdm:
@bors-servo: r+
Thanks for cleaning this up!

PR #115, opened 2016-11-03
Don't make `OsIpcSender` `Sync`

antrik:
`mpsc::Sender` isn't `Sync` -- and the `ipc-channel` equivalent probably
shouldn't be either. Sharing references to senders is usually not a good
idea: they are rather meant to be cloned instead; and they implement
internal sharing to make this efficient and robust -- sharing references
to senders just adds an unnecessary extra layer of sharing on top.

For the `inprocess` back-end, implementing `Sync` was necessitating
use of an `Arc<Mutex<>>`, adding considerable overhead even when not
used -- so this change is an optimisation. It effectively reverts most
of the sender part of 30b2024 (the
receiver part was already backed out in
77469c2 ), except that it obviously
doesn't re-introduce the bogus explicit `Sync` claim.

For the `macos` and `unix` back-ends, the sender implementations were
inherently `Sync`; so we explicitly mark them `!Sync` now, to get a
consistent API.

Also bumping the `ipc-channel` version here, as this is a breaking
change.

(It affects Servo only in one place though, see https://github.com/servo/servo/pull/14048 )

This also comes with a test case to verify that `OsIpcSender` indeed isn't `Sync` any more; and a CI change to activate it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,12 +8,12 @@ os:
   - osx
 
 env:
-  - FEATURES=""
-  - FEATURES="--features force-inprocess"
+  - FEATURES="unstable"
+  - FEATURES="unstable force-inprocess"
 
 notifications:
   webhooks: http://build.servo.org:54856/travis
 
 script:
-  - cargo build $FEATURES
-  - cargo test $FEATURES
+  - cargo build --features "$FEATURES"
+  - cargo test --features "$FEATURES"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ repository = "https://github.com/servo/ipc-channel"
 
 [features]
 force-inprocess = []
+unstable = []
 
 [dependencies]
 bincode = "0.6"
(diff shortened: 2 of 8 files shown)


antrik:
I'm wondering whether maybe it would be better to squash these commits; or some of them at least?...

pcwalton:
Have you tried https://github.com/laumann/compiletest-rs? That strikes me as the simplest way to make sure `OsIpcSender` is `!Sync`.

antrik:
@pcwalton with lots of tears and sweat, I managed to get it working with `compiletest-rs`: https://github.com/servo/ipc-channel/compare/master...antrik:nosync_sender-compiletest

However, `compiletest-rs` actually requires unstable as well -- so I do not really consider it a win. In fact I don't like `compiletest-rs` very much: it's virtually undocumented (thus the tears and sweat); it has annoying quirks (`LD_LIBRARY_PATH` needs to be set explicitly, which is normally not necessary with `cargo` nowadays); and the indirection -- with separate test files in a separate directory executed by a separate test runner -- is not very elegant IMHO. I'd much rather keep my original approach...

PR #119, opened 2016-11-10
Retry macOS select operations instead of panicking

jdm:
#99 made the calculation for buffer sizes for receives work correctly, but there appears to be no guarantee that the size will remain the same when we retry. This can lead to a panic in the current code (as observed in https://github.com/servo/servo/pull/14146#issuecomment-259599900). The test passes if we retry the operation until it succeeds, which matches other code I've found in the wild like https://github.com/openmach/openmach/blob/a185c84f7f5052b7b9f5531ab3612aaf88d3802e/libmach/flick_mach3mig_client.c#L88 .

### DIFF
--- a/src/platform/macos/mod.rs
+++ b/src/platform/macos/mod.rs
@@ -577,29 +577,32 @@ fn select(port: mach_port_t, blocking_mode: BlockingMode)
                                  timeout,
                                  MACH_PORT_NULL) {
             MACH_RCV_TOO_LARGE => {
-                // the actual size gets written into msgh_size by the kernel
-                let max_trailer_size = mem::size_of::<mach_sys::mach_msg_max_trailer_t>() as mach_sys::mach_msg_size_t;
-                let actual_size = (*message).header.msgh_size + max_trailer_size;
-                allocated_buffer = Some(libc::malloc(actual_size as size_t));
-                setup_receive_buffer(slice::from_raw_parts_mut(
-                                        allocated_buffer.unwrap() as *mut u8,
-                                        actual_size as usize),
-                                     port);
-                message = allocated_buffer.unwrap() as *mut Message;
-                match mach_sys::mach_msg(message as *mut _,
-                                         flags,
-                                         0,
-                                         actual_size,
-                                         port,
-                                         timeout,
-                                         MACH_PORT_NULL) {
-                    MACH_MSG_SUCCESS => {},
-                    MACH_RCV_TOO_LARGE => {
-                        panic!("message was bigger than we were told");
-                    }
-                    os_result => {
-                        libc::free(allocated_buffer.unwrap() as *mut _);
-                        return Err(MachError(os_result))
+                loop {
+                    // the actual size gets written into msgh_size by the kernel
+                    let max_trailer_size = mem::size_of::<mach_sys::mach_msg_max_trailer_t>() as mach_sys::mach_msg_size_t;
+                    let actual_size = (*message).header.msgh_size + max_trailer_size;
+                    allocated_buffer = Some(libc::malloc(actual_size as size_t));
+                    setup_receive_buffer(slice::from_raw_parts_mut(
+                                            allocated_buffer.unwrap() as *mut u8,
+                                            actual_size as usize),
+                                         port);
+                    message = allocated_buffer.unwrap() as *mut Message;
+                    match mach_sys::mach_msg(message as *mut _,
+                                             flags,
+                                             0,
+                                             actual_size,
+                                             port,
+                                             timeout,
+                                             MACH_PORT_NULL) {
+                        MACH_MSG_SUCCESS => break,
+                        MACH_RCV_TOO_LARGE => {
+                            libc::free(allocated_buffer.unwrap() as *mut _);
+                            continue;
+                        }
+                        os_result => {
+                            libc::free(allocated_buffer.unwrap() as *mut _);
+                            return Err(MachError(os_result))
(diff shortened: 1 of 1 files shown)


jdm:
cc @jrmuizel

antrik:
@jdm do you have some idea what actually triggers this situation? Or, more specifically, how we could test for this in the `ipc-channel` test suite?...

jdm:
I'm trying to see if I can come up with an isolated testcase as we speak.

jdm:
Going to dump my investigations here:
When I add prinlns to the code that allocates a larger buffer (and right before panicking), I see this output in a non-panicking run:
```
provided 4096, got 12520
provided 4096, got 12084
provided 4096, got 5972
provided 4096, got 11892
provided 4096, got 12024
provided 4096, got 4440
provided 4096, got 169808
provided 4096, got 5956
provided 4096, got 5956
provided 4096, got 5956
```
A panicking run shows me this:
```
provided 4096, got 12520
provided 4096, got 12084
provided 4096, got 5972
provided 4096, got 11892
provided 4096, got 12024
provided 4096, got 4440
provided 4512, got 5956
message was bigger than we were told (thread <unnamed>, at /Users/jdm/src/ipc-channel/src/platform/macos/mod.rs:600)
```
Interestingly, 169808 / 5956 = 28.

jrmuizel:
What size is the message after it fails the first time?

jdm:
That's the "got N".

jdm:
More interesting results from adding a println to the send implementation when the message size exceeds 4096 (the default stack buffer size used by recv):
Successful run:
```
sending 12520 on port 28931
provided 4096, got 12520 on port 5891
sending 12084 on port 9223
provided 4096, got 12084 on port 9223
sending 5972 on port 8711
provided 4096, got 5972 on port 5891
sending 11892 on port 7427
provided 4096, got 11892 on port 29187
sending 12024 on port 148227
provided 4096, got 12024 on port 5891
sending 4440 on port 148227
provided 4096, got 4440 on port 5891
sending 5956 on port 8711
provided 4096, got 5956 on port 5891
sending 5952 on port 8711
provided 4096, got 5952 on port 5891
sending 169808 on port 6147
provided 4096, got 169808 on port 6147
sending 5956 on port 8711
provided 4096, got 5956 on port 5891
sending 5956 on port 8711
provided 4096, got 5956 on port 5891
```
Panicking run:
```
sending 12520 on port 28947
provided 4096, got 12520 on port 5891
sending 12084 on port 9223
provided 4096, got 12084 on port 9223
sending 5972 on port 8711
provided 4096, got 5972 on port 5891
sending 11892 on port 7427
provided 4096, got 11892 on port 30723
sending 12024 on port 47363
provided 4096, got 12024 on port 5891
sending 5956 on port 8711
sending 4440 on port 47363
provided 4096, got 4440 on port 5891
provided 4512, got 5956 on port 5891
message was bigger than we were told (thread <unnamed>, at /Users/jdm/src/ipc-channel/src/platform/macos/mod.rs:603)
```
Suspicious that the sizes involved in the final two sends are the same sizes observed in the erroneous case.

jdm:
More interesting information - the port number showing the issue in the previous run corresponds to an IpcReceiverSet (which is only used in the resource thread to select over the public and private network channels).

antrik:
@jdm I'm a bit confused now: does this patch solve the problem or not?...

jdm:
@antrik The panic does not occur. I no longer believe this is the right solution though, since we may be silently dropping a message.

antrik:
@jdm since we are using `MACH_RCV_LARGE`, this shouldn't ever drop a message according to my reading of http://web.mit.edu/darwin/src/modules/xnu/osfmk/man/mach_msg.html . Rather, the message remains queued, and we can reattempt to get it.

The problem with port sets is probably that when several ports in the set have messages queued, there is apparently no guarantee that on the next attempt we will get us the same one as on the last fail... And if the new one needs a larger buffer than the previous one, we fail again. However, since we grow the buffer each time an attempt fails, we will succeed ultimately.

Of course this is just guesswork... Too bad my attempt at creating a test case for this failed to expose the problem :-(

(BTW, another thing I found in the doc is that apparently we *do* get back the actual trailer size as well on a failed attempt... Though considering the relatively small sizes involved, trying precise allocation here is probably not worth the effort.)

jdm:
My readings agreed with you (and I noticed the trailer size bit as well), but it had not occurred to me that port sets may return a different message each time. I'll see if I can rustle up any further documentation about port sets.

jdm:
Haven't found any supporting documentation yet, so I've started reading the [source](https://github.com/opensource-apple/xnu/tree/10.11/osfmk/ipc). That being said, I realized a complication with the theory - the code in question is an IpcReceiverSet, but it's selecting over the public and private resource thread receivers, while only the public channel is ever used by the test. Naively, it's surprising that a different message could be retrieved, given that the same port from the set would be selected every time.

antrik:
@jdm how shall we proceed with this? Do you want to further investigate why this is happening? Do you want to try coming up with a minimal test case? Or shall we just merge it as is, hoping it really fixes whatever the actual problem is, and it won't come up again?...

jdm:
I've spent more time looking for ideas on and off since this comment. I don't have any good ones. I think we should merge this.

jrmuizel:
This should get a better comment describing how we don't really understand what's going on.

jdm:
I have added a more detailed commit message detailing the erroneous situation.

Issue #120, opened 2016-11-11
benchmark tests: bench_size does not compile


danlrobertson:
After 20ed7d2472977f0eb69a524d1f0b37b09f54d31d [`bench_size`](https://github.com/servo/ipc-channel/blob/master/benches/bench.rs#L30-L75) needs to be updated as it currently fails to compile at [line 40](https://github.com/servo/ipc-channel/blob/master/benches/bench.rs#L40) because `Cell<T>` doesn't implement `Sync`.

PR #121, opened 2016-11-11
Benchmarks: wrap the IpcSender in a mutex

danlrobertson:
After e6ea8390f6f423d3c5def98cb1fdf936a6a2f2ac `IpcSenders` are not `Sync`
and therefore the implicit capture in the closure will cause compile
time failures. If the size is greater than the result of
`get_max_fragment_size` wrap the `IpcSender` in a `Mutex`.

This is what I used to get benchmarks working again for my tests of #94 

Fixes: #120

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -35,12 +35,13 @@ fn bench_size(b: &mut test::Bencher, size: usize) {
     let wait_rx = Mutex::new(wait_rx);
 
     if size > platform::OsIpcSender::get_max_fragment_size() {
+        let safe_tx = Mutex::new(tx);
         b.iter(|| {
             crossbeam::scope(|scope| {
                 scope.spawn(|| {
                     let wait_rx = wait_rx.lock().unwrap();
                     for _ in 0..ITERATIONS {
-                        tx.send(&data, vec![], vec![]).unwrap();
+                        safe_tx.lock().unwrap().send(&data, vec![], vec![]).unwrap();
                         if ITERATIONS > 1 {
                             // Prevent beginning of the next send
                             // from overlapping with receive of last fragment,


emilio:
could we clone the sender instead? Seems like locking kind of would affect the benchmarks right?

danlrobertson:
Good question, but AFAIK no we cannot. I tried that first and didn't get anywhere. Then I noticed that `wait_rx` was a `Mutex<OsIpcReceiver>` and that is when I did the same with `tx`.

antrik:
What @emilio said: the whole *point* of https://github.com/servo/ipc-channel/pull/115 is that senders should be cloned, rather than sharing references across threads... This also allows getting rid of `crossbeam`; just moving the cloned receivers into a "simple" thread closure instead.

(When I wrote the benchmark, I didn't have a good understanding of the interactions yet; and IIRC the `unix` sender wasn't using an internal `Arc<>` yet, to make the cloning cheap... I did have a variant using an explicit `Arc<>` around sender, though -- but at the time, sharing borrowed references through `crossbeam` seemed more elegant to me...)

@danlrobertson if you are not sure how to do this, I can whip up a PR myself.

antrik:
BTW, depending on the platform, cloning the `Arc<>` might actually not be cheaper than locking a mutex... Either is probably negligible though against the cost of spawning a thread.

danlrobertson:
Yup. Feel free to close this PR if you already have an implementation of this. I just wanted to post what I had to get benchmarks working again

antrik:
Here it is: https://github.com/servo/ipc-channel/pull/122

I actually made two patches: the first one is a minimal fix (hint: the trick is in the mysterious `let tx = tx` line -- this is necessary to enforce moving the binding into the closure, i.e. capturing by value, rather than capturing a reference); while the second does away with crossbeam.

danlrobertson:
superseded by #122

PR #123, opened 2016-11-14
platform/test: Test receiver set with concurrent transfers of varying sizes

antrik:
If I'm reading @jdm's analysis in
https://github.com/servo/ipc-channel/pull/119 correctly, this should
hopefully be able to reproduce the problem discussed there.


antrik:
So that didn't work out :-(

antrik:
Folded this into https://github.com/servo/ipc-channel/pull/124 now, since it fails to demonstrate what it was supposed to -- but it should be a useful test in general...

PR #127, opened 2016-12-02
Avoid unnecessary FD cloning

antrik:
This makes a bunch of changes to how FDs are handled in the Unix back-end: most notably, adding inner mutability for the FD in `OsIpcReceiver` by wrapping it in a `Cell<>`, just like the other back-ends do -- thus enabling us to avoid unnecessary `dup()` calls when passing around FDs, without introducing potential FD leaks or premature closing.

This also contains a couple of related cleanups and minor fixes. Furthermore, it adds benchmarks for FD passing, to track the effect of the optimisations; and a bunch of improvements to the existing benchmarks for receiver set handling.

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -304,17 +304,35 @@ mod ipc {
             });
         }
 
-        #[bench]
         // Benchmark performance of removing closed receivers from set.
         // This also includes the time for adding receivers,
         // as there is no way to measure the removing in isolation.
-        fn add_and_remove_closed_receivers(b: &mut test::Bencher) -> () {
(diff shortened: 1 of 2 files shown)


antrik:
Updated to not include the "stolen" benchmark commit, which landed in https://github.com/servo/ipc-channel/pull/94 now.

emilio:
This looks good to me. Haven't reviewed the benchmarks ultra-carefully, but I think they're sound.

@bors-servo r+

emilio:
BTW, awesome commit separation and description, made reviewing this
straight-forward :)

PR #129, opened 2016-12-29
Add books to documentation link

fitzgen:
To test if highfive works.


metajack:
That seemed successful.

PR #130, opened 2017-01-01
platform/unix: Use `reserve_exact()` in `recv()`

antrik:
Since we know exactly at this point how large the message will be, there
is no point in allowing the allocator to reserve more.

While this doesn't actually show any measurable difference here, it
still seems useful to express the constraint explicitly in the code --
to account for possible differences in allocators, for example; and
possibly for clarity?


antrik:
To be honest, this is mostly to stop wasting time pondering whether it's worthwhile submitting a PR for that or not ;-)

jdm:
Seems harmless :)
@bors-servo: r+

PR #108, opened 2016-10-05
Implement ipc-channel on Windows

vvuk:
This implementation uses named pipes on Windows, with auto-generated uuid
names.  It takes advantage of DuplicateHandle to clone handles to target
processes when sending handles cross-process.  Shared memory is implemented
using anonymous file mappings.  select() and friends are implemented using
IO Completion Ports.  I learned a lot about Windows IO doing this, more than I wanted to :)

All the current tests pass.  I have not tried actually running servo with multiprocess ipc -- that's next, but I wanted to get this up for review first.  There's a branch (win32-unsquashed) that has a large sequence of commits that led to this, in case that's interesting to anyone... it contains a couple of abandoned approaches early on though.

PR #132, opened 2017-01-16
Avoid using `include!()` for platform modules

antrik:
Using `include!()` apparently breaks debug info, according to 
https://github.com/servo/ipc-channel/pull/108#discussion_r85535672 -- so
avoid it by going back to per-backend modules.

(Extracted from https://github.com/servo/ipc-channel/pull/108 , since
it's unrelated to the actual purpose of that PR.)

This PR also contains a couple of more or less related cleanups.

### DIFF
--- a/src/platform/inprocess/mod.rs
+++ b/src/platform/inprocess/mod.rs
@@ -21,7 +21,7 @@ use std::mem;
 use std::usize;
 use uuid::Uuid;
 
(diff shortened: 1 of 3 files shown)


jdm:
@bors-servo: retry
* https://github.com/servo/ipc-channel/issues/133

PR #134, opened 2017-01-17
Error handling fixes

antrik:
Some fixes and cleanups for "recursive I/O" and related error handling.


emilio:
Looks great, some comments:

 * r=me on the first commit, I don't know how we could trigger that anyway.
 * r=me on the second commit, with `s/serialisation/serialization` in the commit message.
 * r=me on the third commit, with a question: Do we want to support slightly old rustc versions? Because if so we might want to stick to `try!` (not a big deal IMO, but worth thinking about).
 * And r=me on the last commit, with the same typo corrected, and a comment about why we don't handle the error immediately in the code so it doesn't slip through if someone refactors that in the future.

Thanks for doing this! :)

antrik:
@emilio added a comment regarding the error handling.

I don't think there is any policy about only using features that have been stable for a long time. (This would actually prevent fixing warnings about no longer needed feature gates?) I believe we already use the syntax in various Servo repositories; and I'd be surprised if it's not already being used in various external dependencies as well...

Not changing the spelling of "serialisation", since it's not actually a mistake, as I mentioned on IRC :-)

Issue #131, opened 2017-01-13
mio regressed using a debugger with ipc-channel


jdm:
I ended up doing a build of Servo using ipc-channel master, and discovered that hitting a breakpoint in a debugger caused a ton of "System call interrupted" panics from ipc-channel after resuming. We're going to need to address this before we can use any revision past 357abb967bc4c703af5d4a9f95e2958317e9e8a1 in Servo.

jdm:
cc @danlrobertson

danlrobertson:
You mention `mio` in the issue, but `mio` was not used in the commit mentioned. Did you use `git bisect` to determine that 357abb9 was at fault? If it was the introduction of `mio` that is causing this, the at fault commit would be f6cefc5650e5cca57629f3a9610c20dcc0289fa7.

jdm:
f6cefc5650e5cca57629f3a9610c20dcc0289fa7 makes the panics appear, and be882855182a40e1cbfe60418cd8f87f3278346b makes them disappear.

jdm:
(those are the two most recent commits according to `git log --topo-order`)

jdm:
My original comment could have been written more clearly; I did not mean to indicate that 357abb9 was the commit at fault.

danlrobertson:
Thanks that's super helpful! I'll look into it ASAP.

jdm:
Did you figure anything out, @danlrobertson?

danlrobertson:
I did some work on this last night, and after some digging around carllerche/mio#494 and https://github.com/servo/ipc-channel/blob/master/src/platform/unix/mod.rs#L475-L478 look like likely culprits. Tonight I'm going to try ignoring `EINTR`.

nox:
Can we unstuck this to progress on the servo bump?

danlrobertson:
If you need it ASAP (today) I can pull the tests out and submit a separate PR for that... Otherwise I've got to do some work on that to make it work on Mac. I'd rather keep it as one PR but I totally understand if you need it ASAP

nox:
Beginning of next week is fine too.

PR #137, opened 2017-02-02
Ignore EINTR on unix::OsIpcReceiverSet::select

danlrobertson:
Ignore `EINTR` on Poll::poll in `unix::OsIpcReceiverSet::select`. Returning
`EINTR` as an error on select causes panics in servo.

Resolves: servo/ipc-channel#131

### DIFF
--- a/src/platform/test.rs
+++ b/src/platform/test.rs
@@ -16,6 +16,8 @@ use std::thread;
 
 use platform::{OsIpcSender, OsIpcOneShotServer};
 #[cfg(not(any(feature = "force-inprocess", target_os = "windows", target_os = "android")))]
+use libc::{kill, SIGSTOP, SIGCONT};
+#[cfg(not(any(feature = "force-inprocess", target_os = "windows", target_os = "android")))]
 use test::{fork, Wait};
 
 #[test]
@@ -409,6 +411,44 @@ fn receiver_set() {
             }
         }
     }
+}
+
+#[test]
+#[cfg(not(any(feature = "force-inprocess", target_os = "windows", target_os = "android")))]
+fn receiver_set_eintr() {
+    let (server, name) = OsIpcOneShotServer::new().unwrap();
+    let child_pid = unsafe {
+        fork(|| {
+            let (tx0, rx0) = platform::channel().unwrap();
+            let mut rx_set = OsIpcReceiverSet::new().unwrap();
+            let rx_id = rx_set.add(rx0).unwrap();
+            // Let the parent know we're ready
+            let tx1 = OsIpcSender::connect(name).unwrap();
+            tx1.send(b" Ready! ", vec![OsIpcChannel::Sender(tx0)], vec![]).unwrap();
+            // Send the result of the select back to the parent
+            let result = rx_set.select().unwrap();
+            let mut result_iter = result.into_iter();
+            let (id, data, _, _) = result_iter.next().unwrap().unwrap();
+            assert_eq!(rx_id, id);
+            assert_eq!(b"Test".as_ref(), &*data);
+            assert!(result_iter.next().is_none());
+            tx1.send(b"Success!", vec![], vec![]).unwrap();
+        })
+    };
+    // Wait until the child is ready
+    let (server, res, mut channels, _) = server.accept().unwrap();
+    assert!(res == b" Ready! ");
+    let tx1 = channels.first_mut().unwrap().to_sender();
+    unsafe {
+        kill(child_pid, SIGSTOP);
+        thread::sleep(Duration::from_millis(42));
+        kill(child_pid, SIGCONT);
+    }
+    // The interrupt shouldn't affect the following send
+    tx1.send(b"Test", vec![], vec![]).unwrap();
+    let (res, _, _) = server.recv().unwrap();
+    assert!(res == b"Success!");
+    child_pid.wait();
 }
 
 #[test]
--- a/src/platform/unix/mod.rs
+++ b/src/platform/unix/mod.rs
@@ -19,7 +19,7 @@ use std::collections::HashMap;
 use std::ffi::{CStr, CString};
 use std::fmt::{self, Debug, Formatter};
 use std::hash::BuildHasherDefault;
-use std::io::Error;
+use std::io::{Error, ErrorKind};
 use std::marker::PhantomData;
 use std::mem;
 use std::ops::Deref;
@@ -472,10 +472,19 @@ impl OsIpcReceiverSet {
 
     pub fn select(&mut self) -> Result<Vec<OsIpcSelectionResult>,UnixError> {
         let mut selection_results = Vec::new();
-        match self.poll.poll(&mut self.events, None) {
-            Ok(sz) if sz > 0 => {},
-            _ => { return Err(UnixError::last()); }
-        };
+        let mut num_events = 0;
+        while num_events == 0 {
+            match self.poll.poll(&mut self.events, None) {
+                Ok(sz) => {
+                    num_events = sz;
+                },
+                Err(ref e) => {
+                    if e.kind() != ErrorKind::Interrupted {
+                        return Err(UnixError::last());
+                    }
+                }
+            }
+        }
 
         for evt in self.events.iter() {
             let evt_token = evt.token();


src/platform/unix/mod.rs
```diff
@@ -472,7 +472,12 @@ impl OsIpcReceiverSet {
         let mut selection_results = Vec::new();
         match self.poll.poll(&mut self.events, None) {
             Ok(sz) if sz > 0 => {},
-            _ => { return Err(UnixError::last()); }
+            Err(ref e) if e.kind() != ErrorKind::Interrupted => {
+                return Ok(selection_results);
```
danlrobertson: Debated about what to return here. I think a empty vec makes the most sense.

src/platform/unix/mod.rs
```diff
@@ -472,7 +472,12 @@ impl OsIpcReceiverSet {
         let mut selection_results = Vec::new();
         match self.poll.poll(&mut self.events, None) {
             Ok(sz) if sz > 0 => {},
-            _ => { return Err(UnixError::last()); }
+            Err(ref e) if e.kind() != ErrorKind::Interrupted => {
```
antrik: Why `!=`?

src/platform/unix/mod.rs
antrik: Well, the canonical way of handling `EINTR` is using a retry loop.

In most cases it's *probably* not strictly necessary to handle it here, since typical users calling `select()` in a loop and iterating over the result each time will generally just work -- not sure though we should rely on this, as we never return an empty result otherwise as far as I can tell. On a more conceptual level, it seems less ideal to pass control back to the caller unnecessarily -- better just handle it transparently I'd say.

src/platform/unix/mod.rs
danlrobertson: Good catch. I changed the control flow right before committing and forgot to change this.

antrik:
While the solution is probably fine, what I'd like to understand is why this only came up with `mio`, and wasn't a problem with `poll()`?

src/platform/unix/mod.rs
danlrobertson: Good point... I like the retry better. The update will include this

src/platform/unix/mod.rs
antrik: I wonder whether it would be realistic to add a test case for this? :-)

danlrobertson: Another good point. This one is definitely difficult to reproduce as a unit test. I made my best attempt at it with b03e9db3a2ac87ec40dec3072eda0fa56e62a99c. Parts of it feel a bit hacky, like the "kill sled", but it does cause older versions to panic. It also caught a bug in the code! Hence the new changes to `OsIpcReceiverSet::select`

danlrobertson:
> While the solution is probably fine, what I'd like to understand is why this only came up with mio, and wasn't a problem with poll()?

Agreed. This is really interesting. I'm not really sure that I'll be able to definitively answer this. It may have been just a timing issue? Pre-mio versions do fail the unit test I just added, which causes me to lean towards the timing theory.

src/platform/unix/mod.rs
```diff
@@ -471,8 +471,18 @@ impl OsIpcReceiverSet {
...
-            Ok(sz) if sz > 0 => {},
-            _ => { return Err(UnixError::last()); }
+            Ok(sz) => {
+                if sz < 1 {
+                    return self.select();
+                }
+            },
+            Err(ref e) => {
+                if e.kind() == ErrorKind::Interrupted {
+                    return self.select();
```
antrik: Sneaky ;-)

I think I'd actually be more comfortable with an explicit loop -- not sure how others feel about this...

src/platform/test.rs
```diff
@@ -413,6 +413,31 @@ fn receiver_set() {
...
+#[test]
+#[cfg(not(any(feature = "force-inprocess", target_os = "windows", target_os = "android")))]
+fn receiver_set_eintr() {
+    let (tx0, rx0) = platform::channel().unwrap();
+    let mut rx_set = OsIpcReceiverSet::new().unwrap();
+    rx_set.add(rx0).unwrap();
+    let parent_pid = unsafe {
+        libc::getpid()
+    };
+    unsafe {
```
antrik: Not sure it's a good idea to have consecutive `unsafe` blocks... Generally we just use a larger enclosing block, if there are no (non-trivial) safe operations in between.

src/platform/test.rs
```diff
@@ -413,6 +413,31 @@ fn receiver_set() {
...
+        libc::signal(libc::SIGTERM,
+                     sigterm_handler as fn(libc::c_int) as *mut libc::c_void as libc::sighandler_t);
+    }
+    let _ = unsafe { fork(|| {
```
antrik: No need to explicitly ignore the return value here.

BTW, I wonder whether you could send the signal the other way instead, so only the child process is affected, avoiding problematic interaction with other tests? (Admittedly, this would complicate the code, as the child would have to report back the result I guess...)

src/platform/test.rs
```diff
@@ -413,6 +413,31 @@ fn receiver_set() {
...
+        for _ in 0..10000 {
+            libc::kill(parent_pid, libc::SIGTERM);
+        }
+        tx0.send(b"A", vec![], vec![]).unwrap();
+    })};
+    for _ in rx_set.select().unwrap() {}
```
antrik: Not sure how useful it is to iterate the results without actually checking them... I'd suggest invoking `next()` to get the first result, like some other `receiver_set()` tests are already doing. (Possibly followed by another `next()` to assert that it's `None`.)

src/platform/unix/mod.rs
jdm: I agree.

antrik:
I was thinking of a slightly different test: rather then sending `SIGTERM` and catching it, you could just send `SIGSTOP` followed by `SIGSTART` (after a short delay), like this:

```rust
libc::kill(parent_pid, libc::SIGSTOP);
thread::sleep(Duration::from_millis(3));
libc::kill(parent_pid, libc::SIGCONT);
```

(Possibly with another wait before the first signal -- though it doesn't seem necessary in my testing...)

AIUI, that's more or less what happens when running in a debugger. Unlike the `SIGTERM` approach, it appears to be fully reliable in triggering the failure with mio, even without any looping. It never triggers with the `poll()` implementation though, so indeed this seems to make a difference.

(The only way I got a failure with the old implementation when running multi-threaded is by using a humongous number of `SIGTERM` iterations... It's much easier to trigger with just one thread, but still needs many iterations.)

Another option I considered is using `SIGALRM`. The advantage would be that it doesn't necessarily need a separate process -- though I'm not sure how it interacts with threads; so it might actually still need looping or a separate process for a reliable result. (I haven't actually tested this approach.)

BTW, did you see that the test timed out on the MacOS builder? I wonder whether it actually hangs there for some reason -- or sending many signals just takes very long on MacOS...

src/platform/unix/mod.rs
antrik: Indeed this is not even a real tail call (as @mbrubeck pointed out on IRC), since the `Vec` has a destructor -- so even in optimised builds this clutters the stack... Not much of an issue here, as it's a rare exceptional situation, so the worst thing I guess is the bloated backtrace -- but it's a good thing to keep in mind in general I guess :-)

src/platform/unix/mod.rs
```diff
@@ -471,8 +471,18 @@ impl OsIpcReceiverSet {
     pub fn select(&mut self) -> Result<Vec<OsIpcSelectionResult>,UnixError> {
         let mut selection_results = Vec::new();
         match self.poll.poll(&mut self.events, None) {
-            Ok(sz) if sz > 0 => {},
-            _ => { return Err(UnixError::last()); }
+            Ok(sz) => {
+                if sz < 1 {
+                    return self.select();
```
danlrobertson: In testing `poll` always returned `0` after each call that returned `EINTR`. I need to do some more testing and debugging around this.

src/platform/unix/mod.rs
danlrobertson: Good call. I replaced the recursive calls with a while loop.

src/platform/test.rs
danlrobertson: >  I wonder whether you could send the signal the other way instead

Overall, I like sending the signal to the child better. It could make debugging a little harder if an unexpected panic occurs in the `select`. Overall much better though.

src/platform/test.rs
danlrobertson: Good point

danlrobertson:
> I was thinking of a slightly different test: rather then sending SIGTERM and catching it, you could just send SIGSTOP followed by SIGSTART (after a short delay), like this:

I like this a lot better. The current implementation uses this.

> BTW, did you see that the test timed out on the MacOS builder? I wonder whether it actually hangs there for some reason -- or sending many signals just takes very long on MacOS...

I did notice this, and it does appear to still be happening. I'll try to reproduce it locally on my mac.

> (The only way I got a failure with the old implementation when running multi-threaded is by using a humongous number of SIGTERM iterations... It's much easier to trigger with just one thread, but still needs many iterations.)

I also found this to be true.

> Another option I considered is using SIGALRM. The advantage would be that it doesn't necessarily need a separate process -- though I'm not sure how it interacts with threads

Hadn't really thought about using `SIGALRM`. The current implementation doesn't use it, but I'll continue to look into the best options for producing a reproducible test case.

src/platform/test.rs
```diff
@@ -414,6 +414,55 @@ fn receiver_set() {
...
+    let child_pid = unsafe {
+        fork(|| {
+            // Let the parent know we're ready
+            tx0.send(b"B", vec![], vec![]).unwrap();
+            // Send the result of the select back to the parent
+            match rx_set.select() {
+                Ok(_) => tx0.send(b"\x00", vec![], vec![]).unwrap(),
+                Err(e) => {
+                    let err = format!("{:?}", e);
+                    tx0.send(err.into_bytes().as_slice(), vec![], vec![]).unwrap();
```
antrik: You can just use `as_bytes()` instead of `into_bytes().as_slice()`.

Note that this is not *exactly* the same, since `into_bytes()` consumes the original string, and the resulting temporary byte string is dropped at the end of the expression; while `as_bytes()` keeps the original string in place. In this case (and many others), this doesn't really make a difference though :-)

src/platform/test.rs
```diff
@@ -414,6 +414,55 @@ fn receiver_set() {
...
+        let (_, res, _, _) = result.unwrap();
+        assert!(res.len() > 0);
+        if res[0] != 0 && res.len() != 1 {
+            let ret_string = String::from_utf8(res).unwrap();
+            panic!(ret_string);
+        }
+    }
+    unsafe {
+        // reap the child
+        libc::waitpid(child_pid, &mut stat, 0);
```
antrik: Why not use the `child_pid.wait()`, like other tests are doing?

src/platform/test.rs
```diff
@@ -414,6 +414,55 @@ fn receiver_set() {
...
+    let mut rx_set = OsIpcReceiverSet::new().unwrap();
+    let mut result_set = OsIpcReceiverSet::new().unwrap();
+    rx_set.add(rx1).unwrap();
+    let child_pid = unsafe {
+        fork(|| {
+            // Let the parent know we're ready
+            tx0.send(b"B", vec![], vec![]).unwrap();
+            // Send the result of the select back to the parent
+            match rx_set.select() {
+                Ok(_) => tx0.send(b"\x00", vec![], vec![]).unwrap(),
```
antrik: You are not checking at all any more whether the `select()` actually yields the expected result?

src/platform/test.rs
```diff
@@ -414,6 +414,55 @@ fn receiver_set() {
...
+        libc::kill(child_pid, libc::SIGSTOP);
+        thread::sleep(Duration::from_millis(42));
+        libc::kill(child_pid, libc::SIGCONT);
+    }
+    // The interrupt shouldn't affect the following send
+    tx1.send(b"B", vec![], vec![]).unwrap();
+    // Receive the results from the child and panic if the the
+    // returned value from select was an error.
+    result_set.add(rx0).unwrap();
+    for result in result_set.select().unwrap() {
```
antrik: I don't see a need to use a set here: this channel is only for getting back the result; the actual select test is done when sending to the child...

src/platform/unix/mod.rs
```diff
@@ -470,10 +470,21 @@ impl OsIpcReceiverSet {
 
     pub fn select(&mut self) -> Result<Vec<OsIpcSelectionResult>,UnixError> {
         let mut selection_results = Vec::new();
-        match self.poll.poll(&mut self.events, None) {
-            Ok(sz) if sz > 0 => {},
-            _ => { return Err(UnixError::last()); }
-        };
+        let mut sz = 0;
```
antrik: Nitpick: I generally prefer more meaningful variable names, especially when they have a non-trivial scope -- which `sz` now has after moving it out of the `match`.

src/platform/unix/mod.rs
```diff
@@ -470,10 +470,21 @@ impl OsIpcReceiverSet {
...
+        while sz <=0 {
```
antrik: Nitpick: missing blank after `<=`.

BTW, since this is `usize`, I'd say it would be less misleading to just check `== 0`?

src/platform/unix/mod.rs
```diff
@@ -470,10 +470,21 @@ impl OsIpcReceiverSet {
...
+            match self.poll.poll(&mut self.events, None) {
+                Ok(num_events) => {
+                    if num_events > 0 {
+                        sz = num_events;
+                    }
```
antrik: There is no need for this `if`: you are checking the same thing in the `while` condition.

(Or you could use a `loop` instead of `while`, and do an explicit `break` when getting results; thus keeping the condition here. Not sure though whether this would be clearer, or less clear...)

danlrobertson:
Cleaned up the test a bit and addressed comments in the review. I also went back and found the minimal example needed to trigger the mac hang. The following should cause us to hang on mac

```rust
#[test]
fn basic() {
    let (tx0, rx0) = platform::channel().unwrap();
    let child_pid = unsafe {
        fork(|| {
            // child panics
            tx0.send(b"\x00", vec![], vec![]).unwrap();
        })
    };
    let (res, _, _) = rx0.recv().unwrap();
    assert!(res == b"\x00");
    child_pid.wait();
}
```

src/platform/test.rs
danlrobertson: Fixed in update

src/platform/test.rs
danlrobertson: Fixed in update

src/platform/test.rs
danlrobertson: Agreed. Fixed in update.

src/platform/test.rs
danlrobertson: Old habits die hard 😄

src/platform/unix/mod.rs
danlrobertson: Good call. I swapped `num_events` and `sz`.

src/platform/unix/mod.rs
danlrobertson: Two good catches. This is now cleaner.

src/platform/test.rs
```diff
@@ -414,6 +414,54 @@ fn receiver_set() {
...
+    let child_pid = unsafe {
+        fork(|| {
+            // Let the parent know we're ready
+            tx0.send(b"\x00", vec![], vec![]).unwrap();
+            // Send the result of the select back to the parent
+            match rx_set.select() {
+                Ok(result) => {
+                    for evt in result {
+                        if let OsIpcSelectionResult::DataReceived(_, data, _, _) = evt {
+                            assert!(data == b"\x00");
```
antrik: I'm sorry to say that the error checking is both overly complicated and incomplete: it doesn't check for wrong number of results (including 0); nor for unexpected `ChannelClosed` instead of `DataReceived`...

As I mentioned before, you can just do `let (id, data, _, _) = result.into_iter().next().unwrap().unwrap()`, like existing test cases.

(In case it's not clear, the first unwrap() is for the `next()` returning `Some(...)`, while the second one is a custom method of `OsIpcSelectionResult`.)

Or being really thorough:
```rust
let mut result_iter = result.into_iter();
let (id, data, _, _) = result_iter.next().unwrap().unwrap();
assert!(id == rx_id);
assert!(data == b"\x00");
assert!(result_iter.next().is_none());
```

antrik:
@danlrobertson the other cross-process tests are using `OsIpcOneShotServer`. I suspect the problem is that raw Mach ports -- unlike FDs -- are not automatically cloned on fork(); so the child never actually gets the sender.

(This is actually quite unsurprising -- we have the problem with raw Mach ports in GNU Hurd...)

src/platform/test.rs
danlrobertson: I tried to follow this, but instead of using asserts I formatted strings to send back to the parent so that we get informative error messages if we hit one of these.

danlrobertson:
> I suspect the problem is that raw Mach ports -- unlike FDs -- are not automatically cloned on fork(); so the child never actually gets the sender.

Interesting. I changed the cfg for the test so it doesn't run on mac.

src/platform/test.rs
antrik: Well, I see your point  -- but if you really want to go down this road, you'd have to match *all* the result values manually in the child process, rather than using any asserts/unwraps etc. Or maybe you could use something with `set_hook()` and/or `catch_unwind()`, which would allow sticking with the usual assert/unwrap stuff -- but that seems tricky to get right as well.

To be honest, I'm not sure either is worth the effort. Just send back the correct result I'd say, and panic the child process otherwise -- like the other multi-threaded/multi-process tests do. Running into a test failure is not the common case; when someone does, they  kinda just have to figure out that they need to run the tests with `-- --nocapture` to see the additional error output... Not ideal, but good enough for the tests.

antrik:
@danlrobertson well, it shouldn't be too hard to change the test to use `OsIpcOneShotServer` so it works with the MacOS back-end as well? I think it would be worthwhile.

Though I just realised that there is also another problem there: any messages sent at the platform level will be padded to a multiple of 4 bytes on MacOS -- so you need to avoid odd lengths, or truncate the results, like other platform tests do...

(Or maybe you could consider implementing this as a top-level test instead of a platform test? Automatic serialisation of the messages would avoid such pitfalls, while otherwise not changing much in this case I think; and I'd don't see any obvious obstacles from the top of my head...)

src/platform/test.rs
danlrobertson: Ah! Another good point... I'll just stick with the assert!

danlrobertson:
Updated the unit test to use `OsIpcOneShotServer`, removed some unneeded clutter from the test, and extended everything sent to a multiple of 4 bytes.

src/platform/test.rs
```diff
@@ -412,6 +414,40 @@ fn receiver_set() {
...
+        fork(|| {
+            // Let the parent know we're ready
+            let tx0 = OsIpcSender::connect(name).unwrap();
+            tx0.send(b"\x00\x00\x00\x00", vec![], vec![]).unwrap();
+            // Send the result of the select back to the parent
+            let result = rx_set.select().unwrap();
+            let mut result_iter = result.into_iter();
+            let (id, data, _, _) = result_iter.next().unwrap().unwrap();
+            assert_eq!(rx_id, id);
+            assert_eq!(vec![0, 0, 0, 0], data);
```
antrik: Nitpick: it's rather inconsistent to use a vector here, while you use a byte string slice in `send()`... You could use `assert_eq!(b"\x00\x00\x00\x00".as_ref().to_owned(), data)` or `assert_eq!(b"\x00\x00\x00\x00".as_ref(), &*data)`.

(And while at it, I'd say `[0; 4]` would be more readable than `b"\x00\x00\x00\x00"`, if you want to stick with the 0s. Probably better though to use byte string slices of four actual characters instead -- preferably different one in each `send()`, to tell them apart...)

src/platform/test.rs
```diff
@@ -412,6 +414,40 @@ fn receiver_set() {
 }
 
 #[test]
+#[cfg(not(any(feature = "force-inprocess", target_os = "windows", target_os = "android")))]
+fn receiver_set_eintr() {
+    let (server, name) = OsIpcOneShotServer::new().unwrap();
+    let (tx1, rx1) = platform::channel().unwrap();
+    let mut rx_set = OsIpcReceiverSet::new().unwrap();
```
danlrobertson: > create the main channel and receiver set in the child process, and send tx1 to the parent process through tx0 along with the "ready" message

Good point. The update does this.

src/platform/test.rs
danlrobertson: `assert_eq!(b"\x00\x00\x00\x00".as_ref(), &*data)` I like this best. The update uses this and uses more descriptive strings.

src/platform/test.rs
```diff
@@ -412,6 +414,40 @@ fn receiver_set() {
...
+            let tx0 = OsIpcSender::connect(name).unwrap();
+            tx0.send(b"\x00\x00\x00\x00", vec![], vec![]).unwrap();
+            // Send the result of the select back to the parent
+            let result = rx_set.select().unwrap();
+            let mut result_iter = result.into_iter();
+            let (id, data, _, _) = result_iter.next().unwrap().unwrap();
+            assert_eq!(rx_id, id);
+            assert_eq!(vec![0, 0, 0, 0], data);
+            assert!(result_iter.next().is_none());
+        })
```
danlrobertson: Your right, my bad... I wasn't thinking when I removed the last send :smile:

src/platform/test.rs
```diff
@@ -412,6 +414,47 @@ fn receiver_set() {
 }
 
 #[test]
+#[cfg(not(any(feature = "force-inprocess", target_os = "windows", target_os = "android")))]
+fn receiver_set_eintr() {
+    let (start_server, start_name) = OsIpcOneShotServer::new().unwrap();
+    let (end_server, end_name) = OsIpcOneShotServer::new().unwrap();
```
antrik: Why are you using a separate server (channel) for the final result, rather than just sending the success message over the first channel as well?

Note that this isn't really *wrong*; so I wouldn't mind merging it as is -- it just makes the test more complicated than necessary...

src/platform/test.rs
```diff
@@ -412,6 +414,47 @@ fn receiver_set() {
...
+            assert_eq!(b"Test".as_ref(), &*data);
+            assert!(result_iter.next().is_none());
+            let tx2 = OsIpcSender::connect(end_name).unwrap();
+            tx2.send(b"Success!", vec![], vec![]).unwrap();
+        })
+    };
+    // Wait until the child is ready
+    let (_, res, mut channels, _) = start_server.accept().unwrap();
+    assert!(res == b" Ready! ");
+    assert!(channels.len() == 1);
```
antrik: I don't think that last assert is really necessary: as this receive is not really part of the stuff we are testing here (rather just needed for the test harness), there is no need to be too thorough here. If we don't get the channel we require for the test, the next `unwrap()` will panic anyway.

src/platform/test.rs
danlrobertson: > Why are you using a separate server (channel) for the final result, rather than just sending the success message over the first channel as well?

Sorry, just hadn't used `OneShotServers` before and didn't read the code close enough. Updated in the latest changes.

src/platform/test.rs
danlrobertson: True, this is a complicated enough test, so IMO any removal of unnecessary asserts or code is a win.

PR #139, opened 2017-02-08
tests: `fork()`: Automatically exit child process after closure returns

antrik:
Rather than panicking, just exit the process cleanly.

The idea with this was that the closure is never supposed to return,
since chaos would ensue otherwise, as the child process and the parent
process would continue running the rest of the test suite.

However, the existing `unreachable!()` panic that was supposed to
enforce this, effectively just terminated the process anyway. The panic
message wouldn't even show up unless using `--nocapture`. So we can just
as well make this official, so callers are no longer required to exit
the process explicitly. This actually makes use of `fork()` more
ergonomic.

### DIFF
--- a/src/platform/test.rs
+++ b/src/platform/test.rs
@@ -14,8 +14,6 @@ use std::sync::Arc;
 use std::time::{Duration, Instant};
 use std::thread;
 
-#[cfg(not(any(feature = "force-inprocess", target_os = "windows", target_os = "android")))]
-use libc;
 use platform::{OsIpcSender, OsIpcOneShotServer};
 #[cfg(not(any(feature = "force-inprocess", target_os = "windows", target_os = "android")))]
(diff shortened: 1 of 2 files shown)


metajack:
The only case where this might be different is that panics may be trapped in the application and thus not cause an exit. Is that not a likely scenario?

antrik:
@metajack that's an interesting point, since the test runner *does* trap panics AIUI? I haven't seen this cause any trouble though when there is a panic in a sub-process... (Aside from the parent sometimes hanging waiting for an answer until the test timeout kicks in -- but that's OK I guess...) Maybe that's because a forked process only clones the thread that called `fork()`, so the thread panic causes the process to terminate anyway? Not sure.

At any rate, more explicitly terminating the process is actually more correct in this regard.

Issue #125, opened 2016-11-19
Use memfd for IPCSharedMemory on Linux


Manishearth:
Currently on linux IPCSharedMemory is backed by a file in `/tmp`, which is often not tempfs, and ends up doing unnecessary disk IO.

If possible, we should use [`memfd_create()`](http://man7.org/linux/man-pages/man2/memfd_create.2.html) to take an existing memory buffer and make it into a file descriptor (which we can then share).

cc @antrik @pcwalton

antrik:
@Manishearth this has been on my personal to-do list for a while...

I'm a bit ambivalent about actually using `memfd` though: while it's an elegant concept, it's Linux-only for now. As a first step, we should rather aim for a portable implementation based on the POSIX shared memory API ( http://man7.org/linux/man-pages/man7/shm_overview.7.html ) I'd say.

Manishearth:
Works for me. Patrick had mentioned he preferred memfd over shmem though IIRC.

danlrobertson:
This can be closed since #140 has been merged.

emilio:
Yay, thanks for working on this @danlrobertson! :)

PR #140, opened 2017-02-09
Use memfd_create and shm_open instead of mkstemp

danlrobertson:
Remove the use of `mkstemp` which creates a file in `/tmp`. Instead use the following

 - `linux` - [`memfd_create`](http://man7.org/linux/man-pages/man2/memfd_create.2.html)
 - `freebsd` - [`shm_open`](https://www.freebsd.org/cgi/man.cgi?query=shm_open&sektion=2)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,23 @@
 language: rust
 sudo: false
-rust:
-  - nightly
 
-os:
-  - linux
-  - osx
-
-env:
-  - FEATURES="unstable"
-  - FEATURES="unstable force-inprocess"
+matrix:
+  include:
+    - os: linux
+      rust: nightly
+      env: FEATURES="unstable"
+    - os: linux
+      rust: nightly
+      env: FEATURES="unstable force-inprocess"
+    - os: linux
+      rust: nightly
+      env: FEATURES="unstable memfd"
+    - os: osx
+      rust: nightly
+      env: FEATURES="unstable"
+    - os: osx
+      rust: nightly
+      env: FEATURES="unstable force-inprocess"
 
 notifications:
   webhooks: http://build.servo.org:54856/travis
--- a/src/platform/unix/mod.rs
+++ b/src/platform/unix/mod.rs
@@ -11,7 +11,7 @@ use bincode;
 use fnv::FnvHasher;
 use libc::{self, MAP_FAILED, MAP_SHARED, PROT_READ, PROT_WRITE, SOCK_SEQPACKET, SOL_SOCKET};
 use libc::{SO_LINGER, S_IFMT, S_IFSOCK, c_char, c_int, c_void, getsockopt};
-use libc::{iovec, mkstemp, mode_t, msghdr, off_t, recvmsg, sendmsg};
+use libc::{iovec, mode_t, msghdr, off_t, recvmsg, sendmsg};
 use libc::{setsockopt, size_t, sockaddr, sockaddr_un, socketpair, socklen_t, sa_family_t};
 use std::cell::Cell;
 use std::cmp;
@@ -26,6 +26,8 @@ use std::ops::Deref;
 use std::ptr;
 use std::slice;
 use std::sync::Arc;
+use std::sync::atomic::{AtomicUsize, Ordering, ATOMIC_USIZE_INIT};
+use std::time::UNIX_EPOCH;
 use std::thread;
 use mio::unix::EventedFd;
 use mio::{Poll, Token, Events, Ready, PollOpt};
@@ -40,12 +42,6 @@ const SCM_RIGHTS: c_int = 0x01;
 // is not the size we are actually allowed to use...
 // Empirically, we have to deduct 32 bytes from that.
 const RESERVED_SIZE: usize = 32;
-
-#[cfg(target_os="android")]
-const TEMP_FILE_TEMPLATE: &'static str = "/sdcard/servo/ipc-channel-shared-memory.XXXXXX";
-
-#[cfg(not(target_os="android"))]
-const TEMP_FILE_TEMPLATE: &'static str = "/tmp/ipc-channel-shared-memory.XXXXXX";
 
 #[cfg(target_os = "linux")]
 type IovLen = usize;
@@ -71,6 +67,14 @@ lazy_static! {
         tx.get_system_sendbuf_size().expect("Failed to obtain maximum send size for socket")
     };
 }
+
+// The pid of the current process which is used to create unique IDs
+lazy_static! {
+    static ref PID: c_int = unsafe { libc::getpid() };
+}
+
+// A global count used to create unique IDs
+static SHM_COUNT: AtomicUsize = ATOMIC_USIZE_INIT;
 
 pub fn channel() -> Result<(OsIpcSender, OsIpcReceiver),UnixError> {
     let mut results = [0, 0];
@@ -668,16 +672,14 @@ struct BackingStore {
 
 impl BackingStore {
     pub fn new(length: usize) -> BackingStore {
-        unsafe {
-            let string = CString::new(TEMP_FILE_TEMPLATE).unwrap();
-            let string_buffer = strdup(string.as_ptr());
-            let fd = mkstemp(string_buffer);
-            assert!(fd >= 0);
-            assert!(maybe_unlink(string_buffer) == 0);
-            libc::free(string_buffer as *mut c_void);
-            assert!(libc::ftruncate(fd, length as off_t) == 0);
-            Self::from_fd(fd)
-        }
+        let count = SHM_COUNT.fetch_add(1, Ordering::Relaxed);
+        let timestamp = UNIX_EPOCH.elapsed().unwrap();
+        let name = CString::new(format!("/ipc-channel-shared-memory.{}.{}.{}.{}",
+                                        count, *PID,
+                                        timestamp.as_secs(),
+                                        timestamp.subsec_nanos())).unwrap();
+        let fd = create_shmem(name, length);
+        Self::from_fd(fd)
     }
 
     pub fn from_fd(fd: c_int) -> BackingStore {
@@ -945,18 +947,18 @@ fn recv(fd: c_int, blocking_mode: BlockingMode)
     Ok((main_data_buffer, channels, shared_memory_regions))
 }
 
-#[cfg(target_os="android")]
-fn maybe_unlink(_: *const c_char) -> c_int {
-    // Calling `unlink` on a file stored on an sdcard immediately deletes it.
-    // FIXME: use a better temporary directory than the sdcard via the Java APIs
-    // and threading that value into Servo.
-    // https://code.google.com/p/android/issues/detail?id=19017
-    0
-}
-
-#[cfg(not(target_os="android"))]
-unsafe fn maybe_unlink(c: *const c_char) -> c_int {
-    libc::unlink(c)
+fn create_shmem(name: CString, length: usize) -> c_int {
+    unsafe {
+        // NB: the FreeBSD man page for shm_unlink states that it requires
+        // write permissions, but testing shows that read-write is required.
+        let fd = libc::shm_open(name.as_ptr(),
+                                libc::O_CREAT | libc::O_RDWR | libc::O_EXCL,
+                                0o600);
+        assert!(fd >= 0);
+        assert!(libc::shm_unlink(name.as_ptr()) == 0);
+        assert!(libc::ftruncate(fd, length as off_t) == 0);
+        fd
+    }
 }
 
 struct UnixCmsg {
@@ -1062,7 +1064,6 @@ fn S_ISSOCK(mode: mode_t) -> bool {
 
 extern {
     fn mktemp(template: *mut c_char) -> *mut c_char;
-    fn strdup(string: *const c_char) -> *mut c_char;
 }
 
 #[repr(C)]
--- a/src/platform/unix/mod.rs
+++ b/src/platform/unix/mod.rs
@@ -947,6 +947,7 @@ fn recv(fd: c_int, blocking_mode: BlockingMode)
     Ok((main_data_buffer, channels, shared_memory_regions))
 }
 
+#[cfg(not(all(target_os="linux", feature="memfd")))]
 fn create_shmem(name: CString, length: usize) -> c_int {
     unsafe {
         // NB: the FreeBSD man page for shm_unlink states that it requires
@@ -956,6 +957,16 @@ fn create_shmem(name: CString, length: usize) -> c_int {
                                 0o600);
         assert!(fd >= 0);
         assert!(libc::shm_unlink(name.as_ptr()) == 0);
+        assert!(libc::ftruncate(fd, length as off_t) == 0);
+        fd
+    }
+}
+
+#[cfg(all(feature="memfd", target_os="linux"))]
+fn create_shmem(name: CString, length: usize) -> c_int {
+    unsafe {
+        let fd = memfd_create(name.as_ptr(), 0);
+        assert!(fd >= 0);
         assert!(libc::ftruncate(fd, length as off_t) == 0);
         fd
     }
@@ -1035,6 +1046,11 @@ fn is_socket(fd: c_int) -> bool {
 }
 
 // FFI stuff follows:
+
+#[cfg(all(feature="memfd", target_os="linux"))]
+unsafe fn memfd_create(name: *const c_char, flags: usize) -> c_int {
+    syscall!(MEMFD_CREATE, name, flags) as c_int
+}
 
 #[allow(non_snake_case)]
 fn CMSG_LEN(length: size_t) -> size_t {
--- a/src/platform/unix/mod.rs
+++ b/src/platform/unix/mod.rs
@@ -225,7 +225,7 @@ impl OsIpcSender {
             fds.push(channel.fd());
         }
         for shared_memory_region in shared_memory_regions.iter() {
-            fds.push(shared_memory_region.fd);
+            fds.push(shared_memory_region.store.fd());
         }
 
         // `len` is the total length of the message.
@@ -662,10 +662,69 @@ fn make_socket_lingering(sockfd: c_int) -> Result<(),UnixError> {
     Ok(())
 }
 
+struct BackingStore {
+    fd: c_int
+}
+
+impl BackingStore {
+    pub fn new(length: usize) -> BackingStore {
+        unsafe {
+            let string = CString::new(TEMP_FILE_TEMPLATE).unwrap();
+            let string_buffer = strdup(string.as_ptr());
+            let fd = mkstemp(string_buffer);
+            assert!(fd >= 0);
+            assert!(maybe_unlink(string_buffer) == 0);
+            libc::free(string_buffer as *mut c_void);
+            assert!(libc::ftruncate(fd, length as off_t) == 0);
+            Self::from_fd(fd)
+        }
+    }
+
+    pub fn from_fd(fd: c_int) -> BackingStore {
+        BackingStore {
+            fd: fd,
+        }
+    }
+
+    pub fn fd(&self) -> c_int {
+        self.fd
+    }
+
+    pub unsafe fn map_file(&self, length: Option<size_t>) -> (*mut u8, size_t) {
+        let length = length.unwrap_or_else(|| {
+            let mut st = mem::uninitialized();
+            assert!(libc::fstat(self.fd, &mut st) == 0);
+            st.st_size as size_t
+        });
+        if length == 0 {
+            // This will cause `mmap` to fail, so handle it explicitly.
+            return (ptr::null_mut(), length)
+        }
+        let address = libc::mmap(ptr::null_mut(),
+                                 length,
+                                 PROT_READ | PROT_WRITE,
+                                 MAP_SHARED,
+                                 self.fd,
+                                 0);
+        assert!(address != ptr::null_mut());
+        assert!(address != MAP_FAILED);
+        (address as *mut u8, length)
+    }
+}
+
+impl Drop for BackingStore {
+    fn drop(&mut self) {
+        unsafe {
+            let result = libc::close(self.fd);
+            assert!(thread::panicking() || result == 0);
+        }
+    }
+}
+
 pub struct OsIpcSharedMemory {
     ptr: *mut u8,
     length: usize,
-    fd: c_int,
+    store: BackingStore
 }
 
 unsafe impl Send for OsIpcSharedMemory {}
@@ -678,8 +737,6 @@ impl Drop for OsIpcSharedMemory {
                 let result = libc::munmap(self.ptr as *mut c_void, self.length);
                 assert!(thread::panicking() || result == 0);
             }
-            let result = libc::close(self.fd);
-            assert!(thread::panicking() || result == 0);
         }
     }
 }
@@ -687,9 +744,9 @@ impl Drop for OsIpcSharedMemory {
 impl Clone for OsIpcSharedMemory {
     fn clone(&self) -> OsIpcSharedMemory {
         unsafe {
-            let new_fd = libc::dup(self.fd);
-            let (address, _) = map_file(new_fd, Some(self.length));
-            OsIpcSharedMemory::from_raw_parts(address, self.length, new_fd)
+            let store = BackingStore::from_fd(libc::dup(self.store.fd()));
+            let (address, _) = store.map_file(Some(self.length));
+            OsIpcSharedMemory::from_raw_parts(address, self.length, store)
         }
     }
 }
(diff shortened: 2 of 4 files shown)


src/platform/unix/mod.rs
```diff
@@ -982,6 +970,13 @@ unsafe fn create_shmem(name: *const c_char, length: usize) -> c_int {
     fd
 }
 
+#[cfg(target_os="android")]
+unsafe fn create_shmem(name: *const c_char, length: usize) -> c_int {
+    let fd = ashmem_create_region(name, length);
```
danlrobertson: **NB:** I have not tested the `android` code yet. This is more or less a placeholder for now.

Manishearth:
YES! Thank you!

For mac there are some Mach calls that are really good for this

cc https://github.com/servo/ipc-channel/issues/125

r? @antrik @pcwalton

src/platform/unix/mod.rs
```diff
@@ -1033,6 +1058,11 @@ fn is_socket(fd: c_int) -> bool {
 
 // FFI stuff follows:
 
+#[cfg(target_os="linux")]
+unsafe fn memfd_create(name: *const c_char, flags: usize) -> c_int {
+    syscall!(MEMFD_CREATE, name, flags) as c_int
```
danlrobertson: Kind of less than ideal in my books, but since glibc doesn't provide an wrapper, this is the recommended method. Perhaps we should make `shm_open` the default and this enabled by a feature flag?

src/platform/unix/mod.rs
```diff
@@ -649,10 +649,69 @@ fn make_socket_lingering(sockfd: c_int) -> Result<(),UnixError> {
     Ok(())
 }
 
+struct BackingStore {
```
danlrobertson: In my original design, this played a much bigger role. The `BackingStore` doesn't really do much as is, and could easily be removed. We really rely on the `create_shmem` function.

antrik:
@Manishearth the dedicated MacOS back-end already uses Mach memory objects. This is only about platforms using the Unix back-end.

Manishearth:
>  already uses Mach memory objects.

TIL, nice.

src/platform/unix/mod.rs
antrik: That's odd... Do you know whether `glibc` just doesn't provide a wrapper *yet*, or it doesn't intend to? (And if so, why?)

I don't think it's a reason to disable it by default though. It certainly would be useful however to have a `--force-posix-shm` flag along the lines of `--force-inprocess`, so we can easily test the code locally and in CI.

src/platform/unix/mod.rs
```diff
@@ -873,50 +924,37 @@ fn recv(fd: c_int, blocking_mode: BlockingMode)
...
+    let mut tmp_string = TEMP_FILE_TEMPLATE.to_owned();
+    let addition: String = thread_rng().gen_ascii_chars().take(6).collect();
+    tmp_string.push_str(&addition);
+    CString::new(tmp_string).unwrap()
 }
 
-#[cfg(not(target_os="android"))]
-unsafe fn maybe_unlink(c: *const c_char) -> c_int {
-    libc::unlink(c)
+#[cfg(any(feature="force-shm", target_os="freebsd"))]
```
antrik: Since POSIX SHM is the fallback case, I think it would be better to use `#[cfg(any(feature="force-shm", not(any(target_os="linux", target_os="android")))]`

src/platform/unix/mod.rs
```diff
@@ -873,50 +924,37 @@ fn recv(fd: c_int, blocking_mode: BlockingMode)
...
+#[inline]
```
antrik: Explicit inlining is useful only for functions invoked from other crates.

src/platform/unix/mod.rs
```diff
@@ -649,10 +646,64 @@ fn make_socket_lingering(sockfd: c_int) -> Result<(),UnixError> {
...
+    fd: c_int
+}
+
+impl BackingStore {
+    pub unsafe fn create(length: usize) -> BackingStore {
+        let string = create_filename();
+        let fd = create_shmem(string.as_ptr(), length);
+        BackingStore {
+            fd: fd
+        }
```
antrik: Could use `Self::from_fd()` I guess? Not sure whether that's indeed more readable though...

src/platform/unix/mod.rs
```diff
@@ -873,50 +924,37 @@ fn recv(fd: c_int, blocking_mode: BlockingMode)
...
-#[cfg(target_os="android")]
-fn maybe_unlink(_: *const c_char) -> c_int {
-    // Calling `unlink` on a file stored on an sdcard immediately deletes it.
-    // FIXME: use a better temporary directory than the sdcard via the Java APIs
-    // and threading that value into Servo.
-    // https://code.google.com/p/android/issues/detail?id=19017
-    0
+unsafe fn create_filename() -> CString {
+    let mut tmp_string = TEMP_FILE_TEMPLATE.to_owned();
+    let addition: String = thread_rng().gen_ascii_chars().take(6).collect();
```
antrik: @asajeffrey spent some effort removing use of `thread_rng()` in Servo, since it uses up an FD for each thread...

There is actually no need for this to be random -- so I think a global `lazy_static!` atomic counter would do just fine. It would be less overhead, too.

(In fact I'm not sure there is any value in using unique names with `memfd` at all: the name is only for debugging here; and since we don't keep track of it after initial creation of the FD anyway, I don't think they really help telling the FDs apart...)

Seems more tricky for the POSIX SHM interface though, since the names are global there. Probably need to use the PID along with a counter and timestamp. (We need the timestamp part, since theoretically a process might die before releasing the SHM object, and later another process reusing the PID might collide with that...)

src/platform/unix/mod.rs
```diff
@@ -873,50 +924,37 @@ fn recv(fd: c_int, blocking_mode: BlockingMode)
...
+    CString::new(tmp_string).unwrap()
 }
 
-#[cfg(not(target_os="android"))]
-unsafe fn maybe_unlink(c: *const c_char) -> c_int {
-    libc::unlink(c)
+#[cfg(any(feature="force-shm", target_os="freebsd"))]
+#[inline]
+unsafe fn create_shmem(name: *const c_char, length: usize) -> c_int {
+    let fd = libc::shm_open(name, libc::O_CREAT | libc::O_RDWR, 0660);
```
antrik: Why `660`, not `600`?

(Though in fact it probably should be `000`, since we don't ever open the same object again anyway -- we just clone the existing FD, which is already open with `O_RDWR`...)

Also, `0660` doesn't do what you think it does :-) In Rust, you'd need `0o660`.

Last but not least, it's probably good to add `O_EXCL` just to be safe.

src/platform/unix/mod.rs
```diff
@@ -873,50 +924,37 @@ fn recv(fd: c_int, blocking_mode: BlockingMode)
...
-fn maybe_unlink(_: *const c_char) -> c_int {
-    // Calling `unlink` on a file stored on an sdcard immediately deletes it.
-    // FIXME: use a better temporary directory than the sdcard via the Java APIs
-    // and threading that value into Servo.
-    // https://code.google.com/p/android/issues/detail?id=19017
-    0
+unsafe fn create_filename() -> CString {
+    let mut tmp_string = TEMP_FILE_TEMPLATE.to_owned();
+    let addition: String = thread_rng().gen_ascii_chars().take(6).collect();
+    tmp_string.push_str(&addition);
```
antrik: I think it would be easier to use `format!()` to create the string?

src/platform/unix/mod.rs
antrik: I see that you already did that in fact :-)

antrik:
Just to be clear: it's especially the second commit that is weird, since it refactors code that is obsoleted piece by piece by later commits... It would be much cleaner to do any refactoring necessary on top of the POSIX SHM code as the new base-line, before adding the specific Linux and Android back-ends.

src/platform/unix/mod.rs
```diff
@@ -649,10 +649,69 @@ fn make_socket_lingering(sockfd: c_int) -> Result<(),UnixError> {
     Ok(())
 }
 
+struct BackingStore {
+    fd: c_int
+}
+
+impl BackingStore {
+    pub unsafe fn create(length: usize) -> BackingStore {
```
danlrobertson: Good point. I renamed the function to `new` and moved the code inside of an `unsafe` block. I probably could have moved only parts of the code inside the `unsafe` block, but I thought this looked more readable. I'm okay with changing that though.

src/platform/unix/mod.rs
```diff
@@ -649,10 +649,69 @@ fn make_socket_lingering(sockfd: c_int) -> Result<(),UnixError> {
...
+        libc::free(string_buffer as *mut c_void);
+        assert!(libc::ftruncate(fd, length as off_t) == 0);
+        BackingStore {
+            fd: fd
+        }
+    }
+
+    pub fn from_fd(fd: c_int) -> BackingStore {
+        BackingStore {
+            fd: fd
```
danlrobertson: Updated.

src/platform/unix/mod.rs
danlrobertson: Debated this for a while, but I went ahead and also updated this.

src/platform/unix/mod.rs
danlrobertson: I still have some work to do on this. In the meantime I removed the use of `thread_rng()` in this function and used a combination of the pid and timestamp. To be completely honest, I totally forgot to add the counter 😄 

> Seems more tricky for the POSIX SHM interface though

Yes, AFAIK both of the other platforms should be fine with the `lazy_static`, but it gets a lot trickier with any platforms using the `shm_open` implementation.

src/platform/unix/mod.rs
danlrobertson: Agreed

src/platform/unix/mod.rs
danlrobertson: Good point. Updated.

src/platform/unix/mod.rs
danlrobertson: Updated. 

> Also, 0660 doesn't do what you think it does :-) In Rust, you'd need 0o660.

Old habits die hard

danlrobertson:
> My major qualm is that the changes are in a sub-optimal order: causing a lot of unnecessary code churn, and making review of individual commits very weird...

Good feedback! Reordered the commits and condensed a few commits.

src/platform/unix/mod.rs
antrik: Sorry, my bad: this needs one more closing parenthesis... I'm surprised though that you didn't see the syntax error in local testing :-)

danlrobertson: Wow. Yeah, not sure how that wasn't caught

src/platform/unix/mod.rs
```diff
@@ -873,50 +922,37 @@ fn recv(fd: c_int, blocking_mode: BlockingMode)
...
 }
 
-#[cfg(target_os="android")]
-fn maybe_unlink(_: *const c_char) -> c_int {
-    // Calling `unlink` on a file stored on an sdcard immediately deletes it.
-    // FIXME: use a better temporary directory than the sdcard via the Java APIs
-    // and threading that value into Servo.
-    // https://code.google.com/p/android/issues/detail?id=19017
-    0
+fn create_string() -> CString {
```
antrik: `create_string()` doesn't really have a lot of meaning... Should be something like `create_shm_name()` I'd say -- though actually I'm not convinced it's useful to have a separate function for this at all...

src/platform/unix/mod.rs
```diff
@@ -873,50 +922,37 @@ fn recv(fd: c_int, blocking_mode: BlockingMode)
...
+    let epoc = UNIX_EPOCH.elapsed().unwrap_or(Duration::new(0, 0));
```
antrik: I think you can just `unwrap()` here -- time elapsed since epoch should always be valid.

(Weird that there seems to be no simpler way to obtain this... And also weird that there seems to be no way to obtain a global monotonic timestamp. Doesn't really matter though in this case, as with a precise timestamp along with the PID and a counter, a collision is well-nigh impossible even if system time does jump back...)

`epoc` is a weird variable name, BTW... Just `timestamp` I'd say?

src/platform/unix/mod.rs
```diff
@@ -873,50 +922,37 @@ fn recv(fd: c_int, blocking_mode: BlockingMode)
...
+    let pid = unsafe {
+        libc::getpid()
+    };
```
antrik: For such a simple unsafe expression, I believe it's more customary to put it on a single line.

src/platform/unix/mod.rs
```diff
@@ -873,50 +922,37 @@ fn recv(fd: c_int, blocking_mode: BlockingMode)
...
+    CString::new(&*format!("/ipc-channel-shared-memory.{}.{}.{}",
```
antrik: You don't need the `&*` dance here: `CString::new()` actually *wants* an owned string (or some other byte vector) -- if you turn it into a slice first, this just means it will have to clone it unnecessarily to gain ownership...

src/platform/unix/mod.rs
```diff
@@ -873,50 +922,37 @@ fn recv(fd: c_int, blocking_mode: BlockingMode)
...
+                           pid, epoc.as_secs(), epoc.subsec_nanos())).unwrap()
 }
 
-#[cfg(not(target_os="android"))]
-unsafe fn maybe_unlink(c: *const c_char) -> c_int {
-    libc::unlink(c)
+#[cfg(any(feature="force-posix-shm", not(any(target_os="linux", target_os="android"))))]
+unsafe fn create_shmem(name: *const c_char, length: usize) -> c_int {
```
antrik: This wouldn't need to be `unsafe`, if you pass `name` as a `CString`.

src/platform/unix/mod.rs
```diff
@@ -873,50 +922,37 @@ fn recv(fd: c_int, blocking_mode: BlockingMode)
...
+    let fd = libc::shm_open(name, libc::O_CREAT | libc::O_RDWR | libc::O_EXCL, 0o600);
```
antrik: A quick check seems to confirm that this works with `0o000` as well, as I expected :-)

src/platform/unix/mod.rs
danlrobertson: Tested this today, and it appears to work.

.travis.yml
```diff
@@ -10,6 +10,7 @@ os:
 env:
   - FEATURES="unstable"
   - FEATURES="unstable force-inprocess"
+  - FEATURES="unstable force-posix-shm"
```
danlrobertson: Yeah, I though that was a little strange... I'l look into this

src/platform/unix/mod.rs
danlrobertson: :+1: Nice! I'll change that

src/platform/unix/mod.rs
danlrobertson: Good point! That is definitely a win

src/platform/unix/mod.rs
danlrobertson: > though actually I'm not convinced it's useful to have a separate function for this at all

Originally I was going to generate the name differently for different OSes, but in the end I thought it was probably best to keep it simple. I'll add the name creation to the `BackingStore::create` method.

src/platform/unix/mod.rs
danlrobertson: I'll update this tomorrow

.travis.yml
```diff
@@ -1,15 +1,23 @@
...
+      env: FEATURES="unstable force-inprocess"
+    - os: linux
+      rust: nightly
+      env: FEATURES="unstable force-posix-shm"
+    - os: osx
+      rust: nightly
+      env: FEATURES="unstable"
+    - os: osx
+      rust: nightly
+      env: FEATURES="unstable force-inprocess"
```
danlrobertson: This is a bit more verbose, but it doesn't build with `force-posix-shm` on osx.

danlrobertson:
Updated to address comments from previous review. I also removed the OS specific commit for android. From some continued testing and code review, I realized that the OS specific code for android under the `unix` module is unreachable since the `inprocess` implementation is used. I'll submit an issue and possibly work on a follow-up PR that makes android use the `unix` module.

src/platform/unix/mod.rs
danlrobertson: I just tested this on FreeBSD and unfortunately you can't unlink if you don't open with at least `0o600`

danlrobertson: Sorry `0o200`

src/platform/unix/mod.rs
```diff
@@ -649,10 +647,67 @@ fn make_socket_lingering(sockfd: c_int) -> Result<(),UnixError> {
     Ok(())
 }
 
+struct BackingStore {
+    fd: c_int
+}
+
+impl BackingStore {
+    pub fn new(length: usize) -> BackingStore {
+        let count = GLOBAL_SHM_COUNT.fetch_add(1, Ordering::SeqCst);
```
antrik: While the documentation on `Ordering` is unfortunately pretty useless to mere mortals, after doing some digging, I'm *fairly* confident that `Relaxed` would do here...

src/platform/unix/mod.rs
```diff
@@ -873,50 +928,27 @@ fn recv(fd: c_int, blocking_mode: BlockingMode)
...
-    assert!(maybe_unlink(string_buffer) == 0);
-    libc::free(string_buffer as *mut c_void);
-    assert!(libc::ftruncate(fd, length as off_t) == 0);
-    fd
+#[cfg(any(feature="force-posix-shm", not(target_os="linux")))]
+fn create_shmem(name: CString, length: usize) -> c_int {
+    unsafe {
+        let fd = libc::shm_open(name.as_ptr(),
+                                libc::O_CREAT | libc::O_RDWR | libc::O_EXCL,
+                                0o600);
```
danlrobertson: So the FreeBSD man page says `shm_unlink` needs write privileges, but testing is showing it needs read-write.

src/platform/unix/mod.rs
danlrobertson: Haha. Yeah, I always forget that Relaxed is the same as LLVM's definition of monotomic.

src/platform/unix/mod.rs
```diff
@@ -655,16 +653,14 @@ struct BackingStore {
...
-            libc::free(string_buffer as *mut c_void);
-            assert!(libc::ftruncate(fd, length as off_t) == 0);
-            Self::from_fd(fd)
-        }
+        let count = GLOBAL_SHM_COUNT.fetch_add(1, Ordering::SeqCst);
+        let timestamp = UNIX_EPOCH.elapsed().unwrap();
+        let pid = unsafe { libc::getpid() };
+        let name = CString::new(format!("/ipc-channel-shared-memory.{}.{}.{}",
+                                        count, pid,
+                                        timestamp.as_secs())).unwrap();
```
danlrobertson: > Why did you drop the nanosecond part? It seems perfectly plausible for a process to exec() (thus resetting the counter while keeping the PID), and then create new shared memory objects all within a single second...

I didn't think about the exec scenario. This is definitely the part of the PR that I'm most unsure about at the moment. I guess if we go back to the pid, counter, second, nanosecond combo I'd be happy with it, but I also started to wonder if some sort of random hash might be better...

.travis.yml
danlrobertson: This is now the last commit

src/platform/unix/mod.rs
```diff
@@ -825,7 +880,7 @@ fn recv(fd: c_int, blocking_mode: BlockingMode)
         let bytes_read = try!(cmsg.recv(fd, blocking_mode));
         main_data_buffer.set_len(bytes_read - mem::size_of_val(&total_size));
 
-        let cmsg_fds = cmsg.cmsg_buffer.offset(1) as *const c_int;
+        let cmsg_fds = CMSG_DATA(cmsg.cmsg_buffer) as *const c_int;
```
danlrobertson: Found another spot we weren't using `CMSG_DATA`.

src/platform/unix/mod.rs
antrik: Way to keep programmers on their toes ;-)

Probably good to add a comment for this?

src/platform/unix/mod.rs
antrik: That's not really related to this PR though, is it?...

danlrobertson: I can submit this as a separate one... There are a few other things I've found that I'll want to fix too

danlrobertson: Removed and submitted as servo/ipc-channel#144

src/platform/unix/mod.rs
danlrobertson: Haha so true... Comment added.

src/platform/unix/mod.rs
danlrobertson: Nanosec part was added back in. So now we have `/ipc-channel-shared-memory.<count>.<pid>.<seconds>.<nanoseconds>`

src/platform/unix/mod.rs
```diff
@@ -72,6 +68,8 @@ lazy_static! {
     };
 }
 
+static GLOBAL_SHM_COUNT: AtomicUsize = ATOMIC_USIZE_INIT;
```
antrik: I totally forgot to comment on this: I don't think the `GLOBAL_` prefix is useful. I haven't ever seen something like that used with Rust. (The all-caps already identify it clearly as a global.)

Maybe also good to add a comment that this is used for creating the unique IDs?

src/platform/unix/mod.rs
```diff
@@ -660,10 +658,68 @@ fn make_socket_lingering(sockfd: c_int) -> Result<(),UnixError> {
...
 
+struct BackingStore {
+    fd: c_int
+}
+
+impl BackingStore {
+    pub fn new(length: usize) -> BackingStore {
+        let count = GLOBAL_SHM_COUNT.fetch_add(1, Ordering::Relaxed);
+        let timestamp = UNIX_EPOCH.elapsed().unwrap();
+        let pid = unsafe { libc::getpid() };
```
antrik: I wonder whether it would be useful to put this in a `lazy_static`. I don't know whether this is cached somehow -- but if it isn't, and performs an actual system call every time, this could have a somewhat significant performance impact...

src/platform/unix/mod.rs
antrik: I'm really uncertain myself whether a random hash wouldn't be preferable. I don't know whether other methods of generating random numbers -- such as `weak_rng()`, or perhaps even `StdRng::new()` in a `lazy_static` -- would avoid the FD use problem. If so, it *might* be preferable, since it's probably a tick simpler all in all... Really not sure though. I don't see a strong motivation to change it at this point.

antrik:
@danlrobertson BTW, you might want to rework the PR description -- keep in mind that this ends up in permanent Git history :-)

src/platform/unix/mod.rs
danlrobertson: Added a comment and removed `GLOBAL_`. I left the `SHM_` prefix and  named the `lazy_static` pid `SHM_PID`. I don't really like the `SHM_` prefix, but I wanted to make it a little clearer what it was used for.

src/platform/unix/mod.rs
danlrobertson: Sorry, you mentioned this earlier but I didn't get around to it. Updated per the above comment.

src/platform/unix/mod.rs
antrik: Personally, I don't think `PID` should have a `SHM_` prefix -- while its (currently) used only for this, the PID itself is a general property of the process, not in any way SHM-specific...

I don't feel strongly about it though, and I don't think this warrants another iteration :-)

src/platform/unix/mod.rs
antrik: Nah, I actually only mentioned using `lazy_static` for the counter before -- not realising that an atomic is the more obvious choice for that one...

antrik:
I'd say this is good to go now :-)

After the `mio` adventures, I first need to ask though, whether you tested this with multi-process Servo yet?

danlrobertson:
> After the mio adventures, I first need to ask though, whether you tested this with multi-process Servo yet?

I'm running them now. Would it be better to just submit a "fake" PR and run the servo CI? Just to make sure I don't miss something.

Manishearth:
Yes, that works. Ask someone to run it on try when you do.

danlrobertson:
> Yes, that works.

Awesome! PR submitted and `try`ing now.

danlrobertson:
@antrik I should have brought this up sooner, but `MEMFD_CREATE` wasn't introduced until 3.17. So perhaps we should switch the `force-posix-shm` flag for a `memfd-create` features flag. We could also add a `build.rs` function to check if `MEMFD_CREATE` exists.

antrik:
I don't think submitting "fake PRs" is a sustainable workflow. Unless we can include Servo testing in `ipc-channel` CI (which is probably unrealistic), we just can't expect this do be done on every major change. If we don't trust local testing to be sufficient, probably the best we can do is trying to land risky changes in Servo ASAP, so any problems get exposed in a timely fashion.

antrik:
@danlrobertson Linux 3.17 is a pretty steep requirement indeed...

Build-time feature detection is not sufficient, as the official nightlies for example need to be portable. We could attempt run-time detection -- but that would complicate the code, making it more error-prone and harder to test. I wonder whether it's really worthwhile under these conditions? Maybe it's best to make `memfd` opt-in for now after all...

danlrobertson:
>  If we don't trust local testing to be sufficient, probably the best we can do is trying to land risky changes in Servo ASAP, so any problems get exposed in a timely fashion.

Agreed. Not having a solid definition of a "major" change is complicates things more. IMO keeping `ipc-channel` "up to date" in servo would be ideal, but managing dependencies like `serde` could end up being a bit of a pain. If keeping the `ipc-channel` version at the latest version is too risky, perhaps we could have an `ipc-channel` branch? That would also be a pain to maintain... but it could reduce the risk on introducing a bug into `servo` that stalled other devs work?

> Maybe it's best to make memfd opt-in for now after all

I'm starting to lean towards the `memfd-create` feature as well.

antrik:
Just `memfd` I'd say... The extra verbosity seems more confusing than clarifying in this case.

danlrobertson:
Updated to include a `memfd` feature, rebased on `master`, and updated servo PR.

src/platform/unix/mod.rs
```diff
@@ -886,50 +947,29 @@ fn recv(fd: c_int, blocking_mode: BlockingMode)
...
-unsafe fn create_memory_backing_store(length: usize) -> c_int {
-    let string = CString::new(TEMP_FILE_TEMPLATE).unwrap();
-    let string_buffer = strdup(string.as_ptr());
-    let fd = mkstemp(string_buffer);
-    assert!(fd >= 0);
-    assert!(maybe_unlink(string_buffer) == 0);
-    libc::free(string_buffer as *mut c_void);
-    assert!(libc::ftruncate(fd, length as off_t) == 0);
-    fd
+#[cfg(any(not(feature="memfd"), target_os="freebsd"))]
```
antrik: This should be written in terms of `not(target_os="linux")`. (It's the default implementation, used in all cases unless we have a more specific one...)

In fact it's probably clearer to write it as `#[cfg(not(all(target_os="linux", feature="memfd")))]` -- not least because that way it's obvious that it's complementary to the other one.

(I'd put the `linux` bit before the `memfd` one, since `memfd` is dependant on `linux`... Not really important though.)

danlrobertson:
> The commit message for the last commit needs updating.

Good catch.

> It's the default implementation, used in all cases unless we have a more specific one

Right, my bad... I didn't correctly migrate the `cfg` in the switch from `force-posix-shm` to `memfd`. Good catch.

PR #141, opened 2017-02-11
Cleanup dead code warnings

danlrobertson:
- Add conditional compilation to `incrementor`
 - Remove unused type definitions in test code

### DIFF
--- a/src/platform/mod.rs
+++ b/src/platform/mod.rs
@@ -7,6 +7,7 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
+#[cfg(any(feature = "force-inprocess", not(target_os = "macos")))]
 mod incrementor {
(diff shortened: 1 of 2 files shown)


src/platform/mod.rs
```diff
@@ -7,6 +7,7 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
+#[cfg(any(feature = "force-inprocess", not(target_os = "macos")))]
```
antrik: Gah, my bad: I explicitly removed the conditional in https://github.com/servo/ipc-channel/pull/132/commits/09e22f643ff40ae7562863eb34dfe42431096b4e , as I didn't realise unused modules produce a warning :-(

danlrobertson: Np

PR #146, opened 2017-02-14
Do not panic on unknown readable events on Linux

nox:


nox:
This is crap, see #147 instead.

Issue #133, opened 2017-01-16
Intermittent failure in test::router_big_data on linux


jdm:
```
failures:

---- test::router_big_data stdout ----

	thread 'test::router_big_data' panicked at 'called `Result::unwrap()` on an `Err` value: "SendError(..)"', /buildslave/rust-buildbot/slave/nightly-dist-rustc-linux/build/src/libcore/result.rs:870

failures:

    test::router_big_data
```

It may be related to these other messages that appeared in the log:
```
thread '<unnamed>' panicked at 'Readable event for unknown token: Token(135)', src/platform/unix/mod.rs:503

note: Run with `RUST_BACKTRACE=1` for a backtrace.
```
```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Error { repr: Os { code: 32, message: "Broken pipe" } }', /buildslave/rust-buildbot/slave/nightly-dist-rustc-linux/build/src/libcore/result.rs:870
```

antrik:
@jdm yes, the "unknown token" message is probably the relevant one -- AIUI this shouldn't happen ever, even if there are failures elsewhere...

The error check is in the code introduced by https://github.com/servo/ipc-channel/pull/94 -- though it's not necessarily a regression in the new implementation: IIRC we have had rare intermittents in `ipc-channel` in the past :-( Don't remember exactly which tests, but it very well might have been this one.

Either way, @danlrobertson is probably the most familiar with this part of the code now -- so maybe he has an idea what to do about this? :-)

danlrobertson:
Thanks for bringing this issue up! I think this is because we currently use the file descriptor number as the `Token` [here](https://github.com/servo/ipc-channel/blob/master/src/platform/unix/mod.rs#L460). This could be similar to the same issue fixed with the auto-incrementing id. I see a few possible fixes for this

  - Switch to using `Slab` instead of a `HashMap`.
  - Use an auto-incrementing id. This presents a problem when considering a 32bit arch.

@antrik @jdm thoughts?

antrik:
Ugh... I did investigate this somewhat a while back, but never got around to write down my findings :-(

I'm pretty sure the problem is *not* actually with token collisions.

(If we wanted to avoid using FDs as tokens, that could be done easily: https://github.com/antrik/ipc-channel/commit/deaf8aa1d2127bec5a9b46b758379599b45cdc70 . I'm not sure though it would really be helpful at all...)

[...]

PR #147, opened 2017-02-14
Use slab for unix OsIpcReceiverSet's tokens

danlrobertson:
Use slab as the container for OsIpcReceiverSet's tokens and fds.

Resolves:
 - servo/ipc-channel#133

Related to:
 - servo/servo#15537

### DIFF
--- a/src/platform/unix/mod.rs
+++ b/src/platform/unix/mod.rs
@@ -8,17 +8,14 @@
 // except according to those terms.
 
 use {DeserializeError, SerializeError};
-use fnv::FnvHasher;
 use libc::{self, MAP_FAILED, MAP_SHARED, PROT_READ, PROT_WRITE, SOCK_SEQPACKET, SOL_SOCKET};
 use libc::{SO_LINGER, S_IFMT, S_IFSOCK, c_char, c_int, c_void, getsockopt};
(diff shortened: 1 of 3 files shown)


src/platform/unix/mod.rs
```diff
@@ -457,17 +454,27 @@ impl OsIpcReceiverSet {
...
         };
-        try!(self.poll.register(&io,
-                                fd_token,
-                                Ready::readable(),
-                                PollOpt::level()));
-        self.pollfds.insert(fd_token, poll_entry);
-        Ok(last_index)
+        if self.pollfds.has_available() {
+            let len = self.pollfds.len();
+            self.pollfds.reserve_exact(len << 1);
```
danlrobertson: Not a huge fan of this. Any ideas on how to improve this?

danlrobertson:
It appears that servo/servo#15537 is still failing, which means my theory that it was a reused file descriptor is wrong.

antrik:
Indeed, this is not the problem at all -- see my (late) follow-up on https://github.com/servo/ipc-channel/issues/133

danlrobertson:
Not the actual cause of the problem. Keep `pollfds` as a `HashMap` for now.

PR #148, opened 2017-02-14
platform/unix: Don't return ECONNRESET if dedicated channel is closed

antrik:
So far, when the sender of a dedicated channel in a large transfer was
closed before the receiver got the expected amount of data, `recv()`
returned a "channel closed" status, just as if the main channel was
closed. This is incorrect though, since the main channel is still open,
and the receiver shouldn't be dropped.

Returning `EIO` now instead -- which is a bit of an abuse I guess; but
probably as good a choice as any, as long as we are sticking with Unix
error codes only...

### DIFF
--- a/src/platform/unix/mod.rs
+++ b/src/platform/unix/mod.rs
@@ -881,7 +881,13 @@ fn recv(fd: c_int, blocking_mode: BlockingMode)
         };
 
         if result == 0 {
-            return Err(UnixError(libc::ECONNRESET))
+            // If we stop receiving data before we got the expected amount,
+            // something must have gone seriously wrong.
+            //
+            // EIO doesn't seem like a really good choice --
+            // but probably about the best we can do
+            // without completely overhauling the Error types used here...
+            return Err(UnixError(libc::EIO))
         } else if result < 0 {
             return Err(UnixError::last())
         };


antrik:
I'm not too happy with this: while I *believe* that it is more correct, I don't know yet how to verify that it actually improves anything...

@nox could you do another try run, with this instead of https://github.com/servo/ipc-channel/pull/147 ? Not sure it will change anything -- but I guess it's worth a shot.

nox:
@antrik Sure!

antrik:
To clarify: while it turns out https://github.com/servo/ipc-channel/issues/133 is not affected by this at all, I'm still pretty sure it's more correct this way, and we should commit the change...

danlrobertson:
Is this still valid? It seems like a possible scenario, and I did run a test with servo/ipc-channel#149 and this that ran without issues.

antrik:
It turns out I'm not so sure any more now...

The only scenario I can think of where this would get triggered, is when the sender dies unexpectedly before it finishes the send. Since the main channel normally would also get closed in this case, returning `ECONNRESET` would not actually be wrong... Except that there might be another process still having a sender, in which case it would be definitely wrong. So this probably needs some kind of case handling... Let me ponder it a bit more.

Closing for now to avoid confusion.

PR #149, opened 2017-02-15
Unix: Degregister descriptors that return ECONNRESET

danlrobertson:
Deregister descriptors that return `ECONNRESET` on `recv`.

Resolves: servo/ipc-channel#133

### DIFF
--- a/src/platform/unix/mod.rs
+++ b/src/platform/unix/mod.rs
@@ -500,6 +500,7 @@ impl OsIpcReceiverSet {
                         }
                         Err(err) if err.channel_is_closed() => {
                             self.pollfds.remove(&evt_token).unwrap();
+                            self.poll.deregister(&EventedFd(&poll_entry.fd)).unwrap();
                             unsafe {
                                 libc::close(poll_entry.fd);
                             }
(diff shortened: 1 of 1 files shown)


danlrobertson:
CC @antrik @nox

src/platform/unix/mod.rs
```diff
@@ -500,6 +501,8 @@ impl OsIpcReceiverSet {
                         }
                         Err(err) if err.channel_is_closed() => {
                             self.pollfds.remove(&evt_token).unwrap();
+                            let io = EventedFd(&poll_entry.fd);
+                            self.poll.deregister(&io).unwrap();
```
danlrobertson: After I thought about it and looked at the code more, I realized the real problem is that we don't deregister the descriptor when we hit an error here. Theoretically just deregistering the descriptor here should resolve the issue.

nox:
@antrik Does that look ok to you? It made the try run succeed at least.

nox:
> BTW, any chance we could create an ipc-channel test case for this?

This would be highly valuable, but I wouldn't mind if this lands later.

src/platform/unix/mod.rs
danlrobertson: > This makes total sense now

I know right. After I saw this, it seemed plain as day.

> Explicitly removing the FD should indeed fix this for good. Have you tried whether this alone fixes the failures?

Yes. Just removing the fd causes the tests to pass.

> BTW, is there any particular reason why you introduced the io temporary?

`gdb` if I remember right. Had a breakpoint that printed its value. I can remove it.

nox:
@bors-servo r=antrik

I Git bad.

nox:
@bors-servo retry

I also Homu bad.

PR #153, opened 2017-02-26
Add docs to ipc

danlrobertson:
Add documentation to the structures in the `ipc` module.


danlrobertson:
all the things

danlrobertson:
Opened #154

antrik:
Huh, highfive on strike?...

Anyway, before I get to take a closer look, just some generic remarks for now:

* I don't think there is any value in having separate commits for each type. I'd say only the crate-level documentation should be separate.

* Some of the descriptions essentially just seem to echo what the signatures already tell... The descriptions should try to explain the *purpose* of the various items, not the types they deal with etc.

* I believe I have read that a doc comment should always start with a single proper sentence (ending with a period) briefly describing the item, as this is extracted as the short description... Don't know how strict the parser really is, though.

BTW, I'm not really familiar with rustdoc -- is there some easy way to quickly check how the result looks like?

danlrobertson:
Good comments. I'll work on incorporating them later.

> is there some easy way to quickly check how the result looks like?

`cargo doc` or if you're lazy `cargo doc --open` (essentially `cargo doc; xdg-open ./path/to/doc/index.html`)

danlrobertson:
> I don't think there is any value in having separate commits for each type. I'd say only the crate-level documentation should be separate.

Updated.

> Some of the descriptions essentially just seem to echo what the signatures already tell... The descriptions should try to explain the purpose of the various items, not the types they deal with etc.

I tried to update this, but for most functions I didn't change much. E.g. for `IpcReceiver::try_recv` I left it as "Non-blocking receive". I think we need some sort of documentation here, but I'm not sure that much more would be hlepful.

> I believe I have read that a doc comment should always start with a single proper sentence (ending with a period) briefly describing the item, as this is extracted as the short description... Don't know how strict the parser really is, though.

Yes, it doesn't appear to be too strict. I tried to adhere to this, but for functions like `recv` and `try_recv` I couldn't think of much more to write... I can think about it more and/or add more documentation if it is needed though.

dlrobertson:
Updated to address comments... almost two years later :smile:

Issue #154, opened 2017-02-26
Compile time failure


danlrobertson:
When using the latest nightly rustc and compiling on the master branch I get the following.

```
error[E0282]: type annotations needed
   --> src/ipc.rs:155:42
    |
155 |                     let mut serializer = bincode::Serializer::new(&mut bytes);
    |                         --------------   ^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type for `E`
    |                         |
    |                         consider giving `serializer` a type
error: aborting due to previous error
error: Could not compile `ipc-channel`.
```

jdm:
This is broken on TravisCI, too.

aidanhs:
Quick workaround is to make the dep in cargo read `bincode = "=1.0.0-alpha2"`.

dtolnay:
Until https://github.com/rust-lang/cargo/issues/2222 is resolved, you should always depend on pre-release version using "=1.0.0-alpha2". Semver makes no guarantee about compatibility between alphaN and alphaN+1.

danlrobertson:
This has been resolved

PR #155, opened 2017-02-26
Add endianness to serializers and deserializers

danlrobertson:
Enforce endianness in the creation of our serializers and deserializers
in the ipc module.

Resolves: #154

### DIFF
--- a/src/ipc.rs
+++ b/src/ipc.rs
@@ -11,6 +11,7 @@ use platform::{self, OsIpcChannel, OsIpcReceiver, OsIpcReceiverSet, OsIpcSender}
 use platform::{OsIpcOneShotServer, OsIpcSelectionResult, OsIpcSharedMemory, OsOpaqueIpcChannel};
 
 use bincode::{self, SizeLimit};
+use byteorder::NativeEndian;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use std::cell::RefCell;
 use std::cmp::min;
@@ -152,7 +153,7 @@ impl<T> IpcSender<T> where T: Serialize {
                 let os_ipc_shared_memory_regions;
                 let os_ipc_channels;
                 {
-                    let mut serializer = bincode::Serializer::new(&mut bytes);
+                    let mut serializer = bincode::Serializer::<_, NativeEndian>::new(&mut bytes);
                     data.serialize(&mut serializer)?;
                     os_ipc_channels =
                         mem::replace(&mut *os_ipc_channels_for_serialization.borrow_mut(),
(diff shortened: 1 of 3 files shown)


antrik:
I'm not sure I understand. Is this working around a regression introduced by a change in the compiler? If so, is it really a good idea to work around it, rather than waiting for a fix in the compiler?...

src/ipc.rs
```diff
@@ -152,7 +153,7 @@ impl<T> IpcSender<T> where T: Serialize {
                 let os_ipc_shared_memory_regions;
                 let os_ipc_channels;
                 {
-                    let mut serializer = bincode::Serializer::new(&mut bytes);
+                    let mut serializer = bincode::Serializer::<_, LittleEndian>::new(&mut bytes);
```
antrik: Shouldn't that rather be `NativeEndian`?

(Which is apparently hard-aliased to `LittleEndian` anyway, for whatever reason -- but still it seems a more obvious choice conceptually?...)

jdm:
This is the bincode crate publishing https://github.com/TyOverby/bincode/commit/5cdf2a248bc482da97e3bb091d9d459c1be983ac as a non-semver-compatible version update, presumably (https://github.com/TyOverby/bincode/commit/eebd6a98c3e8fc5c2b8bac2feaf9abf2b1683b1f).

antrik:
Actually, if using `NativeEndian` -- and if that was indeed what the name suggests -- this would in fact be a good change regardless, since there is totally no reason to enforce any particular endianness here, and using the native one could be a tick more efficient on some machines. Otherwise... I'm not sure. Forcing little endian is probably still good, since it's the better choice in *most* cases, and it might not be default.

antrik:
@jdm well, I don't quite understand the interaction between `serde` and `bincode`; but if it's indeed a `bincode` issue (and not compiler-related, as https://github.com/servo/ipc-channel/issues/154 suggests), I would assume it's caused by https://github.com/TyOverby/bincode/pull/103 -- but that was *supposed* to be API-neutral AIUI, in which case the regression should be fixed in `bincode` I guess?...

antrik:
(For the record, even if unintentional, strictly speaking this *is* actually semver-compatible, since semver explicitly states that there are no API guarantees whatsoever for pre-releases... But then again, the Cargo interpretation of semver doesn't match the actual standard anyway.)

src/ipc.rs
danlrobertson: Ah, right... `NativeEndian` is much better.

danlrobertson:
>  I don't quite understand the interaction between serde and bincode; but if it's indeed a bincode issue (and not compiler-related, as #154 suggests)

It is indeed a `bincode`/`serde` issue. I did not describe the issue well. I also do not know `bincode`/`serde` well.

antrik:
Actually, this workaround will work *only* with the breaking API change; so the patch at least has to update the `bincode` version requirement in `Cargo.toml`... However, when the regression is (hopefully) fixed upstream, it won't work with later versions either. (And the broken version presumably gets yanked, so in fact the workaround won't work with *any* version once that happens...)

In other words, this can probably be only addressed in a sane way in `bincode` itself -- which is really the right thing to do anyway. In the free software world, we generally can, and should, fix problems at their source, rather then working around them...

antrik:
See https://github.com/TyOverby/bincode/pull/128

PR #161, opened 2017-03-29
Added sender.send_borrowed(data), to reduce copying

### LINKED ISSUE #156: Reduce copying when sending
This came out of discussions in #138.

The proposal there is to allow sending borrowed data, e.g. a method `send_clone(&data)` with the same semantics as`send(data.clone())` but removing the unnecessary copying.

This deals with borrowing as long as the types match up, but doesn't deal with the case where the types disagree, e.g. sending a `&[&T]` on a channel of type `IpcSender<Vec<T>>`. There is some discussion in #138 about approaches to this.

cc @nox @antrik

asajeffrey:
For compatibility with channels, senders take ownership of their data, and drop it immediately after sending. This PR adds the ability to send borrowed data, including nested data, for example sending a `&(&str,u32)` on a sender expecting a `(String,u32)`.

cc @nox @antrik.

Fixes #156.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ lazy_static = "0.2"
 libc = "0.2.12"
 rand = "0.3"
 serde = "0.9"
+serde_roundtrip = "0.1"
 uuid = {version = "0.4", features = ["v4"]}
 fnv = "1.0.3"
 
--- a/src/ipc.rs
+++ b/src/ipc.rs
@@ -12,6 +12,7 @@ use platform::{OsIpcOneShotServer, OsIpcSelectionResult, OsIpcSharedMemory, OsOp
 
 use bincode;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use serde_roundtrip::RoundTrip;
 use std::cell::RefCell;
 use std::cmp::min;
 use std::fmt::{self, Debug, Formatter};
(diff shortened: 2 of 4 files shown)


jdm:
Someone else who has been more involved in the discussions about #156 should probably review this.

nox:
I still don't think serde-roundtrip is a good idea, so not me.

asajeffrey:
@nox: Just to confirm... Stuff we agree about: having a `RoundTrip<T>` trait, with `#[derive(RoundTrip)]`. Stuff we don't: the `round_trip(&self) -> T` method in `RoundTrip<T>`, and the implementation using `SameDeserialization`. AFAICT this only affects users who implement `RoundTrip` by hand, it's all hidden from people using the custom derive?

@antrik r?

asajeffrey:
@jdm very wise.

Gankro:
FYI, this would let us reuse the allocation of BuiltDisplayLists (~1MB/frame) in webrender.

antrik:
@asajeffrey sorry for slacking on this :-( I did glance at briefly it when you published it, and had a bunch of questions, but didn't manage to write them down at the time...

Please give it another couple of days while I look into it. If I still fail to follow up, feel free to ignore me...

antrik:
@Gankro doesn't Webrender use `BytesChannel` though to avoid serialisation overhead? That one already takes a reference...

asajeffrey:
@antrik?

antrik:
Since my concerns pertain to the general approach, rather than the specific pull request, I deemed it more appropriate to follow up on the discussion at https://github.com/servo/ipc-channel/issues/156 for now.

Gankro:
Sorry I didn't follow up on this. I think I convinced myself that webrender doesn't need this, for the reasons stated by antrik.

asajeffrey:
Closing in favour of experimenting with something like https://github.com/asajeffrey/shared-data

PR #162, opened 2017-05-10
Remove need to truncate mach IPC results

### LINKED ISSUE #83: Trailing data with `bytes_channel` on MacOS

### LINKED ISSUE #103: ipc-channel tests truncate received data to expected size
I don't understand this -- every test truncates the received data to the expected size, e.g.:

https://github.com/servo/ipc-channel/blob/master/src/platform/test.rs#L28
https://github.com/servo/ipc-channel/blob/master/src/platform/test.rs#L139

Is it expected/allowed that you can receive more data than you send, or even a bigger buffer?!

I had a bug where I was accidentally sending data twice, and these tests were happily passing.

jdm:
Mach's IPC requires 4 byte alignment. The code that receives data does not account for that right now; these changes ensure that the returned buffer is the same size as the original buffer that was sent. Fixes #83. Fixes #103.

### DIFF
--- a/src/platform/macos/mod.rs
+++ b/src/platform/macos/mod.rs
@@ -435,12 +435,6 @@ impl OsIpcSender {
             data_dest = data_dest.offset(mem::size_of::<usize>() as isize);
             ptr::copy_nonoverlapping(data.as_ptr(), data_dest, data_size);
 
-            let mut ptr = message as *const u32;
-            let end = (message as *const u8).offset(size as isize) as *const u32;
-            while ptr < end {
-                ptr = ptr.offset(1);
-            }
-
             let os_result = mach_sys::mach_msg(message as *mut _,
                                                MACH_SEND_MSG,
                                                (*message).header.msgh_size,
--- a/src/platform/macos/mod.rs
+++ b/src/platform/macos/mod.rs
@@ -427,8 +427,13 @@ impl OsIpcSender {
             // Zero out the last word for paranoia's sake.
             *((message as *mut u8).offset(size as isize - 4) as *mut u32) = 0;
 
-            let data_dest = shared_memory_descriptor_dest as *mut u8;
-            ptr::copy_nonoverlapping(data.as_ptr(), data_dest, data.len());
+            let mut data_dest = shared_memory_descriptor_dest as *mut u8;
+            let data_size = data.len();
+            let data_size_dest = data_dest as *mut usize;
+            *data_size_dest = data_size;
+
+            data_dest = data_dest.offset(mem::size_of::<usize>() as isize);
+            ptr::copy_nonoverlapping(data.as_ptr(), data_dest, data_size);
 
             let mut ptr = message as *const u32;
             let end = (message as *const u8).offset(size as isize) as *const u32;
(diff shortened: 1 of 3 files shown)


src/platform/macos/mod.rs
```diff
@@ -427,14 +427,13 @@ impl OsIpcSender {
...
+            let data_size_dest = data_dest as *mut usize;
+            *data_size_dest = data_size;
 
-            let mut ptr = message as *const u32;
-            let end = (message as *const u8).offset(size as isize) as *const u32;
-            while ptr < end {
-                ptr = ptr.offset(1);
-            }
+            data_dest = data_dest.offset(mem::size_of::<usize>() as isize);
+            ptr::copy_nonoverlapping(data.as_ptr(), data_dest, data_size);
```
antrik: I can't see the message size being adjusted anywhere for the extra space taken up by the size header -- am I missing something?...

(This kind of pointer manipulation is so damn fragile -- it's actually worse than in C :-( Can't really blame you for that though I guess: probably not much point trying to use slices for this particular bit, without rewriting the rest of function in a more defensive way too...)

src/test.rs
```diff
@@ -371,7 +371,7 @@ fn embedded_bytes_receivers() {
     let (super_tx, super_rx) = ipc::channel().unwrap();
     super_tx.send(sub_tx).unwrap();
     let sub_tx = super_rx.recv().unwrap();
-    let bytes = [1, 2, 3, 4, 5, 6, 7, 8];
+    let bytes = [1, 2, 3, 4, 5, 6, 7];
```
pcwalton: Add a comment linking to this PR saying that the odd number of bytes is intentional.

src/platform/macos/mod.rs
```diff
@@ -836,123 +877,213 @@ impl From<MachError> for bincode::Error {
     }
 }
 
+impl From<kern_return_t> for MachError {
+    fn from(code: kern_return_t) -> MachError {
+        if code == MACH_MSG_SUCCESS {
```
pcwalton: Can this be a `match` statement instead?

jdm:
@antrik Github is showing the commits in the wrong order; they are actually ordered in the way you recommend.

src/platform/macos/mod.rs
jdm: I definitely had code which adjusted the header size. I'm confused by how it disappeared!

jdm:
I am not going to make a separate PR for the error changes. They were useful for diagnosing the issue here, and it's not worth separating them out.

pcwalton:
@bors-servo: r+

🎉

antrik:
@jdm the reason I prefer to submit independently useful changes as separate PRs is that in the (hopefully unlikely) case that the main PR doesn't get accepted for some reason, or turns out problematic and needs to be reverted later, it would be good to have the independent changes in nevertheless. But as I said, it's not really a problem...

src/platform/macos/mod.rs
antrik: To clarify, I do see code doing the adjustment in the receiver -- but not in the sender...

jdm: I understood your original question. I remain confused about how code that I wrote and committed in Message::size_of somehow did not make it to github both times I added it. I have verified that it is now present in the commits in this PR.

PR #164, opened 2017-05-29
Bump serde to 1.0, bincode to 0.8 and bump version

Eijebong:


jdm:
Why are we reverting to an earlier version of bincode?

Eijebong:
Unless I misunderstood something, bincode did 1.0-alpha => 0.8

Might be me though

src/ipc.rs
```diff
@@ -342,16 +342,16 @@ impl OpaqueIpcMessage {
...
             OS_IPC_SHARED_MEMORY_REGIONS_FOR_DESERIALIZATION.with(
                     |os_ipc_shared_memory_regions_for_deserialization| {
                 mem::swap(&mut *os_ipc_channels_for_deserialization.borrow_mut(),
                           &mut self.os_ipc_channels);
                 mem::swap(&mut *os_ipc_shared_memory_regions_for_deserialization.borrow_mut(),
                           &mut self.os_ipc_shared_memory_regions);
-                let mut data = &*self.data;
-                let mut deserializer = bincode::Deserializer::new(&mut data, bincode::Infinite);
+                let data = bincode::read_types::SliceReader::new(&self.data[..]);
+                let mut deserializer = bincode::Deserializer::new(data, bincode::Infinite);
```
antrik: Shouldn't the binding better be called something else than `data` after this change?... (`reader` perhaps? Or maybe `data_reader`?...)

antrik:
@Eijebong note that while it's true that `0.8` is the most recent release,`0.7` and `1.0.0-alpha7` have been released at the same time... Looks to me like they are supposed to be separate branches being developed in parallel?

I don't know how to decide which one is more appropriate to use for us, though.

nox:
Pinging @Gankro too who is the last who has touched bincode stuff. What's the current state of art over there? Why is there `0.8` and still alpha `1.0` versions?

Eijebong:
ping @TyOverby, @Gankro told us to ping you concerning bincode's versioning.

TyOverby:
@Eijebong: bincode 0.8.0 is indeed the most recent release

src/ipc.rs
```diff
@@ -62,12 +62,12 @@ pub fn bytes_channel() -> Result<(IpcBytesSender, IpcBytesReceiver),Error> {
 }
 
 #[derive(Debug)]
-pub struct IpcReceiver<T> where T: Deserialize + Serialize {
+pub struct IpcReceiver<T> where T: for<'de> Deserialize<'de> + Serialize {
```
TyOverby: I'd check with @dtolnay, but I'm pretty sure that since you can't taking advantage of zero-copy parsing, you could replace all of these lifetime bounds with `Deserialize<'static>`.

src/ipc.rs
```diff
@@ -342,16 +342,16 @@ impl OpaqueIpcMessage {
...
             OS_IPC_SHARED_MEMORY_REGIONS_FOR_DESERIALIZATION.with(
                     |os_ipc_shared_memory_regions_for_deserialization| {
                 mem::swap(&mut *os_ipc_channels_for_deserialization.borrow_mut(),
                           &mut self.os_ipc_channels);
                 mem::swap(&mut *os_ipc_shared_memory_regions_for_deserialization.borrow_mut(),
                           &mut self.os_ipc_shared_memory_regions);
-                let mut data = &*self.data;
-                let mut deserializer = bincode::Deserializer::new(&mut data, bincode::Infinite);
+                let reader = bincode::read_types::SliceReader::new(&self.data[..]);
+                let mut deserializer = bincode::Deserializer::new(reader, bincode::Infinite);
```
TyOverby: Is there any reason that you aren't just using `bincode::deserialize` here?

Eijebong: Probably because I didn't know it existed. I'll try with it

Eijebong: Yep, it seems to be working.

src/ipc.rs
Eijebong: Would this really make a difference ?

dtolnay: `Deserialize<'static>` is never what you want. `for<'de> Deserialize<'de>` or `DeserializeOwned` is correct. See https://serde.rs/lifetimes.html#trait-bounds.

TyOverby: @dtolnay: Ah yep, it was DeserializeOwned that I was thinking of.

@Eijebong: It would clean up all of the type signatures.  That's about it though.

PR #166, opened 2017-06-02
[WIP] Add native Windows back-end

antrik:
This picks up from https://github.com/servo/ipc-channel/pull/108

For now, it's just commit refactoring (squashing/splitting) of the original PR -- though the intention is to add a series of follow-up commits next, doing some improvements to the actual code, according to my latest review of the original PR.

Since the implementation is essentially working at this point though, it could also be merged as is, if that turns out necessary.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ env:
     - RUST_BACKTRACE=1
   matrix:
     - FEATURES="unstable"
+    - FEATURES="unstable memfd"
 
 notifications:
   webhooks: http://build.servo.org:54856/travis
--- a/src/platform/test.rs
+++ b/src/platform/test.rs
@@ -177,9 +177,15 @@ fn with_n_fds(n: usize, size: usize) {
     let (super_tx, super_rx) = platform::channel().unwrap();
 
     let data: Vec<u8> = (0..size).map(|i| (i % 251) as u8).collect();
-    super_tx.send(&data[..], sender_fds, vec![]).unwrap();
+    let thread = {
+        let data = data.clone();
+        thread::spawn(move || {
+            super_tx.send(&data[..], sender_fds, vec![]).unwrap();
+        })
+    };
     let (received_data, received_channels, received_shared_memory_regions) =
         super_rx.recv().unwrap();
+    thread.join().unwrap();
 
     assert_eq!(received_data.len(), data.len());
     assert_eq!(&received_data[..], &data[..]);
--- a/src/platform/windows/aliased_cell.rs
+++ b/src/platform/windows/aliased_cell.rs
@@ -1,0 +1,363 @@
+// Copyright 2018 The Servo Project Developers. See the COPYRIGHT
+// file at the top-level directory of this distribution.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+use std::mem::{self, ManuallyDrop};
+use std::thread;
+
+/// Dummy type that panics when its destructor is invoked.
+#[derive(Debug)]
+struct DropBomb();
+
+impl Drop for DropBomb {
+    fn drop(&mut self) {
+        let message = "Trying to drop an AliasedCell, which may still have aliases outstanding.";
+        if thread::panicking() {
+            eprintln!("{}", message);
+        } else {
+            panic!(message);
+        }
+    }
+}
+
+/// A wrapper for unsafely aliased memory locations.
+///
+/// `AliasedCell' makes sure that its inner value
+/// is completely inaccessible from safe code.
+/// Once an `AliasedCell` has been constructed from some value,
+/// until the value is moved out again with the (unsafe) `into_inner()` method,
+/// the only way to access the wrapped value
+/// is through the unsafe `alias_mut()` method.
+///
+/// This is useful for FFI calls that take raw pointers as input,
+/// and hold on to them even after returning control to the caller.
+/// Since Rust's type system is not aware of such aliases,
+/// it cannot provide the usual guarantees about validity of pointers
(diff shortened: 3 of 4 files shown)


antrik:
Note: the Linux test failure is a known intermittent in the test condition (happening under memory pressure), not in the actual `ipc-channel` implementation -- it is on my list to fix when I get around to it...

antrik:
Dropped `truncate()` calls from the newly introduced tests -- missed that during the rebase...

nox:
What's the status on this?

antrik:
@nox every time I get to work on the issues I already know, I find other bits to improve as well... So it's hard to give an estimate. If I don't run into any serious problems, it could be done in a few days -- but I wouldn't count on that...

nox:
No problem @antrik, was just doing a bit of triaging.

reset:
@antrik we'd love a separate windows-stable branch (if this is it, that's ok too) for the Habitat project to consume. We can provide any feedback you need and have a team of Rust programmers ourselves if you need any help. Thanks so much for all the hard work here!

antrik:
@reset I realised that maintaining a "stable" branch will only be easy if I can still rewrite history at will like in the WIP branch... I can add a tag however every time I update it, so you would still have persistent reference points. (Though no proper bisecting between them...) Would that work for you?

reset:
@antrik rewriting history is actually just fine by us! If the branch is rebased against master in a regular enough cadence then we can just add this branch as the source for the cargo dep

antrik:
@reset this doesn't sound like a good idea to me... I'm not sure what exactly it implies to add a branch as the source of a cargo dep -- but it sounds bad in either case: if it means cargo always fetches the actual current state of the branch at build time, that would open the possibility of your build changing or even breaking due to changes in an external repository -- which is kinda against the whole idea of cargo... If on the other hand it means cargo registers the specific commit the branch points to at the time `cargo update` is run, that wouldn't work either, since after pushing a new history for the branch, the previous commits are gone from the repository...

The tags on the other wouldn't change once created; and they would make sure the commits stay around even when the branch changes.

reset:
@antrik right, since you're rewriting history the commit hash will change. A tag would be perfect :)

antrik:
@reset just to be clear: if cargo registers the specific commit (hash) on `cargo update`, it should still be fine to point it to the branch, as long as I remember to add a tag each time I publish, so the commits don't go away later :-)

antrik:
I pushed the first version of the "stable" branch to https://github.com/antrik/ipc-channel/tree/windows-stable ( https://github.com/antrik/ipc-channel.git ), along with tag https://github.com/antrik/ipc-channel/tree/0.8.0_windows1 ( https://github.com/antrik/ipc-channel.git ). Let's hope I didn't break anything yet :-)

I also created a pseudo-PR from that branch at https://github.com/servo/ipc-channel/pull/169 , mostly to have confirmation that the things I push to that branch actually work...

reset:
@antrik I'll test the branch out in a build tonight and get back. Super appreciate it!

nox:
Maintaining a tag for something that is a work in progress sounds like a very bad idea to me.

antrik:
@nox there might be a misunderstanding here: my intention is to create a dedicated tag for every new "stable" snapshot.

nox:
> @nox there might be a misunderstanding here: my intention is to create a dedicated tag for every new "stable" snapshot.

This is what I understood. Tags stay forever. I guess that's fine as long as these tags don't leave your fork.

reset:
@nox you can purge tags on a remote with `git push <remote> :<tag>`. Is there another more permanent thing you're talking about in your dev/release process?

antrik:
So I finally managed to work through all the comments I made on the latest version of the original PR (plus a bunch of others I found along the way...) -- however, in the process I also discovered a pretty fundamental issue with the buffer handling, breaking the contract associated with `unsafe` code blocks... So I still don't really feel comfortable merging this as is :-( However, the fix will be rather invasive, and I'm not sure how long it will take...

For now, I pushed a new version to the `windows-stable` branch, tagged with `0.8.0_windows2`.

jdm:
Thanks for continuing to look into this @antrik!

reset:
@antrik thanks for the update and all the hard work!

antrik:
A little belatedly, I pushed another update to `windows-stable` (tag `0.9.0_windows1`), syncing with the 0.9 master release. (Along with some further refactoring I did in the Windows back-end.)

I had hoped it might end up not being necessary, if I managed to wrap up my rework fast enough... But it turned out to be even more of a rabbit hole than I anticipated.

(On the plus side, I *think* I have a workable design now that I can finish implementing...)

retep998:
>But it turned out to be even more of a rabbit hole than I anticipated.

Dang humans always trying to get in my hole...

Boscop:
Any update on this? :)

antrik:
@Boscop well, the truth is, while working on the big refactor to isolate unsafety, I got stuck on a design choice a couple months ago that I couldn't decide on for a while, and got distracted... I only resumed working on this recently.

There is good news though. For one, I finally learned how to compile-check for a different target, which makes working on this a *lot* easier. (It's trivial actually: just install the target rustlibs, and then `cargo check --target i686-pc-windows-msvc`... Not sure why I didn't realise this before.)

More importantly, I made progress with the refactoring -- for the first time I feel that I'm actually nearing the end...

...though that actually doesn't matter all that much any more: I figured out a way to isolate the unsafety *without* needing much refactoring after all!

I think the refactoring will still be useful, putting all the related unsafe code in one place -- but with the actual unsafety leaking no longer an issue, the refactoring doesn't need to land in the initial PR.

There are still a couple of things I want to wrap up; but nothing very tricky any more I believe :-)

ebkalderon:
Anyone know the current status of this PR? Is no one working on it anymore?

SamuelMarks:
How's this progressing?

ofek:
Any update?

jdm:
#233 has the latest updates.

jdm:
Fixed in #272

SamuelMarks:
Great to see @jdm

PS: Might be worth updating the top-level README.md for this repo also

PR #168, opened 2017-09-06
Add iOS support

NatalyaKovalova:
ability to build ipc-channel for iOS platform have been added in this PR.
 For building could be used:
  `cargo build  --target aarch64-apple-ios`

@larsbergstrom , could you, please, make code review and accept changes?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -7,7 +7,7 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-#![cfg_attr(any(feature = "force-inprocess", target_os = "windows", target_os = "android"),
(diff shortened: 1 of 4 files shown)


antrik:
I wonder, was there any attempt and/or discussion about using the `macos` back-end on iOS, rather than the `inprocess` band-aid? Since iOS is based on the same kernel AIUI, it seems to me there might be a good chance that the `macos` back-end would work here as well...

larsbergstrom:
@antrik I totally agree! We were just taking a similar approach to the initial Servo Windows port - hack and slash to stand up something that initially works, and then build out additional platform feature support (especially around multiprocess/isolation features) as desired.

PR #175, opened 2017-10-12
Update to bincode 0.9

jrmuizel:
bincode 0.9 comes with a reasonably big performance fix for
deserialization. WebRender has already switched.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ipc-channel"
-version = "0.8.0"
+version = "0.9.0"
 description = "A multiprocess drop-in replacement for Rust channels"
 authors = ["The Servo Project Developers"]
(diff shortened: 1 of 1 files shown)


jrmuizel:
@jdm can you do a release too?

jdm:
Please update the crate version, in that case.

jrmuizel:
Done

antrik:
Was there actually some change that could break downstream users: either in the API (I think error codes are the only thing from `bincode` we expose downstream?); or in the binary format? If so, that should be mentioned in the commit message; and if not, AIUI the convention would be only to bump to `0.8.1`, thus avoiding unnecessary churn in downstream users?...

jrmuizel:
@antrik rust-semverver complained about a bunch of error related things changing. I'll include more information in the commit message next time.

PR #177, opened 2017-11-05
test.rs: Fix platform conditionals for `wait()`

antrik:
The conditionals need to match those for `fork()`, since these two are
always used together. They got out of sync in
3e9aee8ead2c16ef15432300bfd189ba23f70989


emilio:
Looks good, thanks!

@bors-servo r+

PR #178, opened 2017-11-06
Use anonymous shared memory on FreeBSD

myfreeweb:
It works in Capsicum capability mode (process sandbox).


src/platform/unix/mod.rs
```diff
@@ -947,11 +947,9 @@ fn recv(fd: c_int, blocking_mode: BlockingMode)
...
 
-#[cfg(not(all(target_os="linux", feature="memfd")))]
+#[cfg(not(any(target_os="freebsd", all(target_os="linux", feature="memfd"))))]
 fn create_shmem(name: CString, length: usize) -> c_int {
     unsafe {
-        // NB: the FreeBSD man page for shm_unlink states that it requires
-        // write permissions, but testing shows that read-write is required.
         let fd = libc::shm_open(name.as_ptr(),
                                 libc::O_CREAT | libc::O_RDWR | libc::O_EXCL,
                                 0o600);
```
dlrobertson: You should be able to decrease this in the linux specific `shm_open`. Last time I tested this, FreeBSD required read-write, but it was not required on linux.

src/platform/unix/mod.rs
```diff
@@ -962,6 +960,20 @@ fn create_shmem(name: CString, length: usize) -> c_int {
     }
 }
 
+#[cfg(target_os="freebsd")]
+fn create_shmem(_name: CString, length: usize) -> c_int {
+    unsafe {
+        // NB: the FreeBSD man page for shm_unlink states that it requires
+        // write permissions, but testing shows that read-write is required.
+        let fd = libc::shm_open(1 as *mut i8, // SHM_ANON
```
dlrobertson: :+1: Could we add this to `libc`

src/platform/unix/mod.rs
myfreeweb: Yeah, I thought about that. It's not just for Linux though, it's for everything else, and I'm not sure where write-only would be ok. Linux has `memfd` anyway…

dlrobertson: Ah, true

src/platform/unix/mod.rs
myfreeweb: Sure — https://github.com/rust-lang/libc/pull/837

src/platform/unix/mod.rs
antrik: `memfd` usage is optional on Linux, though. (IIRC that was decided because `memfd` is a pretty recent addition, so it would fail on older Linux versions...)

As for other systems, it's entirely unclear. Actually, I'm not sure we can even rely on it Linux: while it seems more logical to require only write permissions, neither the Linux nor the POSIX man pages seem to mention anything about this at all... I guess it's safer to stick with the broader permissions -- it's not like it has any real downsides here, aside from being confusing.

Either way, I don't think it's appropriate to remove the comment entirely. Even if the code is no longer used on FreeBSD, it doesn't make the comment wrong -- and we need to keep *some* explanation for why the code is this way...

src/platform/unix/mod.rs
```diff
@@ -962,6 +960,20 @@ fn create_shmem(name: CString, length: usize) -> c_int {
     }
 }
 
+#[cfg(target_os="freebsd")]
+fn create_shmem(_name: CString, length: usize) -> c_int {
```
antrik: This is a bit dirty: since the name construction has side effects, I doubt the compiler is smart enough to realise it can be omitted altogether...

Though to be honest, I'm not at all sure whether it's really worthwhile to refactor the code in order to avoid this -- so I guess that's just me thinking out loud, rather than an actual request to change anything :-)

src/platform/unix/mod.rs
myfreeweb: @antrik well I moved the comment to the `#[cfg(target_os="freebsd")]` section, I did not remove it entirely :)

src/platform/unix/mod.rs
antrik: What's the status on this? Could you perhaps request a new release to get this functionality exposed?...

src/platform/unix/mod.rs
antrik: @myfreeweb I am aware that you moved it -- but since you didn't modify the original code, it's still not valid to just drop the comment entirely... Feel free to change the wording though, if you feel the original comment would be confusing in view of the changed situation.

myfreeweb:
What do you think about pulling this code into a separate tiny crate? This exact functionality is needed in a lot of things (mostly Wayland-related). Would you merge a patch that replaces this code with a dependency?

dlrobertson:
I think that is reasonable. `memfd_create` is only available post linux 3.17 and has no glibc wrapper. It must be called via syscall, so as long as that it is only used when a feature is enabled, I think a separate crate would be reasonable.

myfreeweb:
Done! https://github.com/myfreeweb/shmemfdrs & updated this PR to use it

src/lib.rs
```diff
@@ -26,10 +26,9 @@ extern crate mio;
 #[cfg(all(not(feature = "force-inprocess"), any(target_os = "linux",
                                                 target_os = "freebsd")))]
 extern crate fnv;
-#[cfg(all(feature = "memfd", not(feature = "force-inprocess"),
-          target_os="linux"))]
-#[macro_use]
-extern crate syscall;
+#[cfg(all(not(feature = "force-inprocess"), any(target_os = "linux",
```
dlrobertson: Related to myfreeweb/shmemfdrs#1. The old implementation had a default that was used if the target os was not linux (with memfd) or freebsd.

dlrobertson: The issue is no longer valid, but this still does not allow e.g. openbsd to use the `shmemfdrs` crate

myfreeweb: You mean the hardcoded `any(target_os="linux", target_os="freebsd")` in here? Yeah, I would also like to replace them with… `all(unix, not(any(target_os="macos", target_os="ios")))` I guess?

dlrobertson: Actually, I think you are right again here. I seem to have done the same thing with `extern crate mio` and this is the same as the `cfg` for `mod unix`, so there should probably be a separate issue to possibly modify the `cfg`s.

antrik: This is a bit confusing, since the implementation of the `unix` back-end should work with any Unix-like OS, but the conditional *selecting* the `unix` back-end (in https://github.com/servo/ipc-channel/blob/master/src/platform/mod.rs ) only works on specific platforms :-(

The only real change here as far as I can tell is that the `shmemfdrs` dependency requires introducing (yet another) redundant copy of the `unix` conditional.

(A long-standing issue on my personal ToDo-list is to try to factor out the platform conditionals in a central place, so everything else would only have to test for a custom `platform_unix` conditional, or something along these lines...)

myfreeweb: Yeah. Arguably, it's not *another* copy, it's `syscall`'s copy :D

+1 that it's a separate issue

antrik: @dlrobertson oh, I see you already figured that out :-)

antrik: @myfreeweb nah, `syscall` doesn't have an exact copy of the `unix` conditional, since that one is (and always was) Linux-specific, so it didn't have the `not generic enough` problem. But yeah, separate issue.

antrik:
Looks good to me. I'm not sure though what Servo's policy is for introducing new external dependencies?

@myfreeweb out of curiosity: how and for what purpose does WR need to use this?

@dlrobertson that's a bit tangential: but did we ever discuss checking for `memfd` support at run-time?...

myfreeweb:
WR? WebRender, on its own? Probably doesn't.

Servo needs `SHM_ANON` to run sandboxed on FreeBSD: https://github.com/servo/servo/issues/11625#issuecomment-342243384

dlrobertson:
@antrik I think we discussed it but I think I forgot to create an issue for it and therefore forgot to investigate further.

antrik:
@myfreeweb whoops, I guess I misread "Wayland" as "Webrender"...

So I'm a bit confused now: are there other users for `shmemfdrs` outside of `ipc-channel`?

myfreeweb:
No existing users as the crate was created today :) But potential users like https://github.com/Smithay/wayland-window/issues/14

antrik:
@myfreeweb going by the discussion there, it seems like they actually need a way to select the specific mechanism depending on client requests, rather than an automatic abstraction? So it doesn't seem like this crate will really help there...

The reason I'm bringing this up is because I am somewhat reluctant -- and from what I gathered, other Servo developers are too -- to introduce a new crate for code that is not actually likely to be used outside of Servo. That would just increase maintenance burden.

myfreeweb:
The mechanism being file descriptors or not file descriptors — not different ways of creating file descriptors :)

antrik:
@myfreeweb I was pretty sure it was actually about temporary file based descriptors vs. anonymous descriptors? But then again, I'm not really familiar with the Wayland protocol -- so I guess I might be misreading it...

myfreeweb:
I made [similar changes work in weston](https://github.com/myfreeweb/weston/commit/174d428a4422f3f59a6629cb63be2eb22f1e1f39), so I know something :) Of course no one cares how a file descriptor is made.

[Shared memory file descriptors](https://jan.newmarch.name/Wayland/SharedMemory/) are used to pass software-rendered buffers from the client to the compositor. The primary "other way" is [passing GPU (EGL) buffers](https://jan.newmarch.name/Wayland/EGL/). And a compositor theoretically could support EGL only.

PR #180, opened 2017-11-11
Mio updates

dlrobertson:
`Event::kind` has been deprecated for `Event::readiness`. Replace all
usages of `Event::kind` for `Event::readiness`. Bump mio version.


dlrobertson:
Thanks for the quick review!

PR #181, opened 2017-11-19
cleanup: Use `RangeFrom` rather than custom `Incrementor`

antrik:
Use standard library functionality instead of custom class.

When I reviewed the PR that originally introduced this, it *did* feel
wrong to me to implement such a generic helper class in our code... But
for some reason it didn't occur to me that a simple range iterator is
the obvious answer.


antrik:
cc @dlrobertson

PR #185, opened 2018-01-01
OpenBSD & NetBSD support

worr:
Requires accounting for ancillary data in sending data. 63+ fd tests
are disabled, since this blocks on OpenBSD.

Also requires splitting sends and receives into separate threads during
testing.

### DIFF
--- a/src/platform/test.rs
+++ b/src/platform/test.rs
@@ -94,11 +94,16 @@ fn medium_data() {
     let data: Vec<u8> = (0..65536).map(|i| (i % 251) as u8).collect();
     let data: &[u8] = &data[..];
     let (tx, rx) = platform::channel().unwrap();
-    tx.send(data, vec![], vec![]).unwrap();
+    let thread = thread::spawn(move || {
+        let data: Vec<u8> = (0..65536).map(|i| (i % 251) as u8).collect();
+        let data: &[u8] = &data[..];
+        tx.send(data, vec![], vec![]).unwrap();
+    });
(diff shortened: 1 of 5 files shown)


src/platform/unix/mod.rs
```diff
@@ -38,17 +38,21 @@ const SCM_RIGHTS: c_int = 0x01;
 
 // The value Linux returns for SO_SNDBUF
 // is not the size we are actually allowed to use...
-// Empirically, we have to deduct 32 bytes from that.
-const RESERVED_SIZE: usize = 32;
+// Empirically, we have to deduct 72 bytes from that.
+const RESERVED_SIZE: usize = 72;
```
dlrobertson: Is this the same for `linux`, `freebsd`, `netbsd`, and `openbsd`?

src/platform/test.rs
```diff
@@ -94,11 +94,16 @@ fn medium_data() {
     let data: Vec<u8> = (0..65536).map(|i| (i % 251) as u8).collect();
     let data: &[u8] = &data[..];
     let (tx, rx) = platform::channel().unwrap();
-    tx.send(data, vec![], vec![]).unwrap();
+    let thread = thread::spawn(move || {
```
dlrobertson: why is `thread::spawn` needed here and in other tests?

jdm:
Closing due to lack of response. We're merging basic changes that allow the crate to build on OpenBSD; if there are still changes required for running tests, we can accept a separate PR to address those.

PR #187, opened 2018-02-15
Use sc

dlrobertson:
Use `sc` instead of `syscall`. It is the better maintained fork.


jdm:
@bors-servo r+
thanks!

PR #191, opened 2018-03-07
Port to openbsd

kpcyrd:
These changes are needed to compile ipc-channel on openbsd. ipc-channel itself is already compatible and no changes are needed. I've verified this with a project I'm working on that depends on this patch for openbsd support (the program itself has been tested on openbsd with a patched ipc-channel).


emilio:
Looks fine, thank you very much!

@bors-servo r+

PR #195, opened 2018-05-08
Enable backtraces on CI

antrik:
Make it easier to debug test failures caught by CI.

This also includes two extra commits refactoring `.travis.yml`, which are only related in that they touch the same code -- but since I now learned how the build matrix works, I can just as well do some tweaking while at it... And it's probably not worth the trouble to push this in a separate PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: rust
 sudo: false
 
+env:
+  global:
+    - RUST_BACKTRACE=1
+
 matrix:
   include:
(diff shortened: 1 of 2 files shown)


antrik:
@dlrobertson you might be interested in the `.travis.yml` refactoring commits, since they are related to the change you originally did while introducing the `memfd` feature...

jdm:
@bors-servo r+
I like these improvements. Thanks!

PR #196, opened 2018-05-13
update mktemp to mkdtemp to supress warning mktemp is dangerous

akoserwal:
warning: the use of `mktemp' is dangerous, better use `mkstemp' or `mkdtemp'
 #11

### DIFF
--- a/src/platform/unix/mod.rs
+++ b/src/platform/unix/mod.rs
@@ -596,7 +596,7 @@ impl OsIpcOneShotServer {
             loop {
                 let path_string = CString::new(&b"/tmp/rust-ipc-socket.XXXXXX"[..]).unwrap();
                 path = path_string.as_bytes_with_nul().iter().cloned().collect();
-                if *mktemp(path.as_mut_ptr() as *mut c_char) == 0 {
+                if *mkdtemp(path.as_mut_ptr() as *mut c_char) == 0 {
                     return Err(UnixError::last())
                 }
 
@@ -1077,7 +1077,7 @@ fn S_ISSOCK(mode: mode_t) -> bool {
 }
 
 extern {
-    fn mktemp(template: *mut c_char) -> *mut c_char;
+    fn mkdtemp(template: *mut c_char) -> *mut c_char;
 }
 
 #[repr(C)]


src/platform/unix/mod.rs
```diff
@@ -596,7 +596,7 @@ impl OsIpcOneShotServer {
             loop {
                 let path_string = CString::new(&b"/tmp/rust-ipc-socket.XXXXXX"[..]).unwrap();
                 path = path_string.as_bytes_with_nul().iter().cloned().collect();
-                if *mktemp(path.as_mut_ptr() as *mut c_char) == 0 {
+                if *mkdtemp(path.as_mut_ptr() as *mut c_char) == 0 {
```
dlrobertson: `mkdtemp` creates a directory right? Wouldn't that cause problems with the subsequent `new_sockaddr_un` call?

akoserwal: @dlrobertson: right, it should be mkstemp ?

dlrobertson: We're using `mkstemp` to basically generate a random path. I'm not sure that `mkstemp` will work either since it actually opens the file. I think we have a few options.

1) Create the path some other way (e.g. with a UUID).
2) use mkdtemp and then use some path on top of that.

dlrobertson: As opposed to rolling our own, is there anything keeping us from using [tempdir in the tempfile crate](https://docs.rs/tempfile/3.0.2/tempfile/fn.tempdir.html)?

Then we could just add a `dir` member to the `OsIpcOneShotServer` method that holds the `TempDir` and we can use something like

```rust
    let dir = TempDir::new().unwrap();
    let dir_path = dir.path().join("rust-ipc-socket");
    let path_str = CString::new(dir_path.to_str().unwrap()).unwrap();
    path = path_str.as_bytes_with_nul().iter().cloned().collect();
```

akoserwal: @dlrobertson : getting 6 failures by using tempfile create. same as using uuid/File:create.
 platform::test::cross_process
    platform::test::cross_process_sender_transfer
    platform::test::receiver_set_eintr
    platform::test::server_accept_first
    platform::test::server_connect_first
    test::cross_process_embedded_senders
I probably dig more to understand, why these cases are failing.

dlrobertson: Do you have your branch posted somewhere? `TempDir` implements `Drop`, so it might be closing the directory you're creating.

akoserwal: @dlrobertson :https://github.com/akoserwal/ipc-channel/tree/fix-tempdir-repace-mktemp
When we run cargo test. it runs all test-cases even for different os platforms? I can see cross_process_sender_transfer for target os windows fails because Permission denied writing to file (file:///checkout/src/libstd/env.rs)

akoserwal: @dlrobertson : #198 got it working. all test are passing.

antrik: @dlrobertson you are evil: first you proposed two different valid approaches, but when he implemented one of them, you suddenly ask why he didn't do it another way... ;-)

antrik: @akoserwal you may have figured it out in the meantime; but just to be sure: the issue here is that we do *not* want to actually create a regular file. Rather, the `bind()` system call creates a socket special file -- but we need to find a safe location for `bind()` to put it.

One way to do this is creating a unique and unguessable name up front. (`mktemp()` is problematic, because the names it creates are not sufficiently unique/unguessable. Your approach using `uuid` OTOH should have been fine.)

The other way is to safely create a directory for exclusive use (using `mkdtemp()` or `tempfile::TempDir`), and then use any name inside that directory for the socket. That's what you are doing in the newer code.

dlrobertson: > @dlrobertson you are evil: first you proposed two different valid approaches, but when he implemented one of them, you suddenly ask why he didn't do it another way... ;-)

lol yeah. I didn't think of using `tempfile` early enough. Thought it might make the implementation easier.

PR #197, opened 2018-05-17
Fix notification handling tests and `unix` implementation

antrik:
The first commit attempts to make notification tests more robust. (Hopefully fixing intermittent failures on Linux.) The other commits add some additional related tests, and fix a (minor) issue uncovered by them.

### DIFF
--- a/src/platform/test.rs
+++ b/src/platform/test.rs
@@ -716,6 +716,56 @@ fn no_receiver_notification() {
     }
 }
 
+/// Checks for broken pipe notification when receiver is closed
+/// while there are pending unreceived messages.
+///
+/// This can result in a different error condition
+/// than dropping the receiver before a send is attempted.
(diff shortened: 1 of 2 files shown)


metajack:
Are there no other solutions? Sleeps feel like a bit of a hack.

antrik:
@metajack my assumption was that if it's a thread scheduling issue, a sleep of any duration should make it reliable, as it explicitly hands off scheduling... But it seems my assumption was wrong, as the sleep didn't fix the intermittents.

I have implemented polling in a loop now instead -- let's hope this works better.

In order to verify that the loop indeed handles delayed notifications, I added separate tests forcing a delay -- and uncovered an actual issue in the notification handling on Linux in the process... So the PR grew in scope :-)

jdm:
One of the TravisCI jobs seems displeased: https://travis-ci.org/servo/ipc-channel/jobs/385864860

PR #198, opened 2018-05-20
replaced mktemp with tempfile for supress warning mktemp is dangerous

akoserwal:


akoserwal:
#11

src/platform/unix/mod.rs
```diff
@@ -594,12 +596,18 @@ impl OsIpcOneShotServer {
...
             let mut path: Vec<u8>;
             loop {
-                let path_string = CString::new(&b"/tmp/rust-ipc-socket.XXXXXX"[..]).unwrap();
+                let dir = TempDir::new().unwrap();
+                let dir_path = dir.path().join("rust-ipc-socket");
+                
+                //avoid deleting temp director automatically
+                let _tmp_path = dir.into_path();
+              
+                let mut _tmp_file = tempfile_in(&_tmp_path).unwrap();
```
dlrobertson: Are `_tmp_path` and `_tmp_file` needed? also `dir` doesn't live long enough. `TempDir` implements `Drop`, so it will be closed at the end of this funtion.

akoserwal: yes, test run got stuck. If I remove _tmp_path and _tmp_file.  tempdir is getting dropped before socket bind operation. 
```
test platform::test::sender_transfer ... ok
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: UnixError(2)', libcore/result.rs:945:5
test platform::test::cross_process ... test platform::test::cross_process has been running for over 60 seconds
test platform::test::cross_process_sender_transfer ... test platform::test::cross_process_sender_transfer has been running for over 60 seconds
```
Same case when I remove _tmp_path. and use either `File::create` or `tempfile_in` using tempDir path. 

 ``` // let _tmp_path = dir.into_path();
    let mut _tmp_file = tempfile_in(&dir).unwrap(); ```

dlrobertson: > If I remove _tmp_path and _tmp_file. tempdir is getting dropped before socket bind operation.

Can't you add `dir` to the `OsIpcOneShotServer` so it doesn't get dropped.

src/platform/unix/mod.rs
```diff
@@ -577,6 +579,7 @@ impl OsOpaqueIpcChannel {
 
 pub struct OsIpcOneShotServer {
     fd: c_int,
+    _dir:TempDir,
```
dlrobertson: might want to add a comment as to why this is needed.

src/platform/unix/mod.rs
```diff
@@ -31,6 +31,8 @@ use std::time::UNIX_EPOCH;
 use std::thread;
 use mio::unix::EventedFd;
 use mio::{Poll, Token, Events, Ready, PollOpt};
+use tempfile::{TempDir};
```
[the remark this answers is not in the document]
akoserwal: sure.

antrik:
@akoserwal you seem to have two PRs open for the same thing? Unless I'm missing something, please close one of them...

src/platform/unix/mod.rs
```diff
@@ -593,13 +598,15 @@ impl OsIpcOneShotServer {
         unsafe {
             let fd = libc::socket(libc::AF_UNIX, SOCK_SEQPACKET, 0);
             let mut path: Vec<u8>;
+            let dir = TempDir::new().unwrap();
```
antrik: Please use `tempfile::Builder` (https://docs.rs/tempfile/3.0.2/tempfile/struct.Builder.html), to get a meaningful directory name. (The actual file name can then be shortened to just "socket" I guess.)

src/platform/unix/mod.rs
```diff
@@ -593,13 +598,15 @@ impl OsIpcOneShotServer {
...
+            let dir_path = dir.path().join("rust-ipc-socket");
+            
             loop {
```
antrik: With the new approach, you can drop the loop and the `EINVAL` handling I believe: AIUI we only needed to retry, because using `mktemp()` is not atomic. (Someone else might have taken the name between our `mktemp()` and `bind()` calls.) `tempfile` doesn't have that issue.

Might be prudent though to do this change in a separate commit, just in case... (While all your other changes of course will have to be squashed into one before this can be merged.)

src/platform/unix/mod.rs
```diff
@@ -597,34 +598,26 @@ impl OsIpcOneShotServer {
...
                 return Err(UnixError::last())
             }
 
             Ok((OsIpcOneShotServer {
-                fd: fd, _dir: dir,
-            }, String::from_utf8(CStr::from_ptr(path.as_ptr() as
+                fd: fd,
+                _temp_dir: _temp_dir,
+            }, String::from_utf8(CStr::from_ptr(path_string.as_ptr() as
                                                 *const c_char).to_bytes().to_owned()).unwrap()))
```
antrik: This complicated conversion is not necessary any more: you actually already have the final rust string after the `socket_path.to_str().unwrap()` (before converting it to `CString`) -- you just need to add a `.to_owned()`, and save it in another temporary variable. (Or even chain this conversion right in the assignment of `socket_path`.)

src/platform/unix/mod.rs
```diff
@@ -31,8 +31,7 @@ use std::time::UNIX_EPOCH;
 use std::thread;
 use mio::unix::EventedFd;
 use mio::{Poll, Token, Events, Ready, PollOpt};
-use tempfile::TempDir;
-
+use tempfile::{Builder,TempDir};
```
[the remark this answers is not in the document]
akoserwal: @antrik: No problem, it's imp & I appreciate it. thanks

src/platform/unix/mod.rs
```diff
@@ -592,23 +598,13 @@ impl OsIpcOneShotServer {
...
-                    return Err(UnixError::last())
-                }
-
-                let (sockaddr, len) = new_sockaddr_un(path.as_ptr() as *const c_char);
-                if libc::bind(fd, &sockaddr as *const _ as *const sockaddr, len as socklen_t) == 0 {
-                    break
-                }
+            let temp_dir =  Builder::new().tempdir().unwrap();
+            let socket_path = temp_dir.path().join("socket");
+            let path_string = socket_path.to_str().to_owned().unwrap();
```
antrik: `to_owned()` in this place is useless. My suggestion was to add it *after* the `unwrap()`, to convert the resulting `&str` into a `String`. (Maybe I should have said `to_string()` instead to make it clearer... Not sure whether there is consensus nowadays which one is preferable.)

Though doing the `to_string()` conversion only while returning the result, as you did, is fine too -- i.e. you don't need the extra `to_owned()` or `to_string()` here at all.

src/platform/unix/mod.rs
```diff
@@ -592,23 +598,13 @@ impl OsIpcOneShotServer {
...
 
-                let errno = UnixError::last();
-                if errno.0 != libc::EINVAL {
-                    return Err(errno)
-                }
+            let (sockaddr, len) = new_sockaddr_un(path_string.as_ptr() as *const c_char);
```
antrik: There was a misunderstanding here: you still do need to convert the string to a `CString` before passing it to `new_sockaddr_un`. My point was just that you can grab the string to be returned at the end *before* doing the conversion to `CString`, rather than converting the `CString` back to a standard `String`...

antrik:
BTW, I said to add `closes https://github.com/servo/ipc-channel/issues/11`, rather than just `closes #11`, for a reason: while it looks the same in the GitHub interface, it's *not* the same when looking at the commit message through other means...

src/platform/unix/mod.rs
antrik: No, that's what you had before... Your last version was actually closer.

Sorry if I'm not making myself clear. I was thinking of something like

```rust
    let name = socket_path.to_str().unwrap();
    let name_ffi = CString::new(name).unwrap();
```

(And later return `name.to_string()`.)

Maybe you could even do the `CString` conversion directly in the `new_sockaddr_un()` call, since you do not need the value of `name_ffi` anywhere else...

akoserwal: @antrik : thank you so much for keep patience with me.

src/platform/unix/mod.rs
```diff
@@ -592,23 +598,13 @@ impl OsIpcOneShotServer {
...
-                path = path_string.as_bytes_with_nul().iter().cloned().collect();
-                if *mktemp(path.as_mut_ptr() as *mut c_char) == 0 {
-                    return Err(UnixError::last())
-                }
-
-                let (sockaddr, len) = new_sockaddr_un(path.as_ptr() as *const c_char);
-                if libc::bind(fd, &sockaddr as *const _ as *const sockaddr, len as socklen_t) == 0 {
-                    break
-                }
+            let temp_dir =  Builder::new().tempdir().unwrap();
```
antrik: I just noticed there is a second blank after the `=` here as well...

src/platform/unix/mod.rs
```diff
@@ -617,7 +613,8 @@ impl OsIpcOneShotServer {
 
             Ok((OsIpcOneShotServer {
                 fd: fd,
-            }, String::from_utf8(CStr::from_ptr(path.as_ptr() as
+                _temp_dir: temp_dir,
+            }, String::from_utf8(CStr::from_ptr(path_string.as_ptr() as
                                                 *const c_char).to_bytes().to_owned()).unwrap()))
```
antrik: You didn't remove the back-and-forth conversion here -- but that was the whole point of shuffling around the other conversions/bindings above...

akoserwal: So, this will be just
`path_string.to_string()`
right?

antrik: Indeed :-)

akoserwal: done

antrik:
Looks good now :-)

Please squash the commits into one, and give it a nice commit message, describing what you did and why.

(The title line could read something like "unix: Use `tempfile` crate for creating socket location". How much you write in the long description, depends on how ambitious you feel about this -- I for my part would mention getting rid of `mktemp()` for security concerns and the associated linker warning; briefly explain how we instead safely create an exclusive temporary directory to put the socket in; and how we do not need the loop any more. Not forgetting to point out that as an extra bonus, this takes care of cleanup, since `tempdir` does that automatically... And of course closing the issue.)

Just in case you are unsure how to a good commit message could look like, I suggest reading through a bunch of existing commit messages for inspiration :-)

PR #200, opened 2018-05-30
Allow benchmarks to run to completion on macOS

jdm:
These commits fix a series of panics that I encountered while attempting to run benchmarks on macOS. The benchmarks now complete, as can be observed in #199.

### DIFF
--- a/src/test.rs
+++ b/src/test.rs
@@ -428,3 +428,11 @@ fn test_reentrant() {
     sender.send(null.clone()).unwrap();
     assert_eq!(null, receiver.recv().unwrap());
 }
+
+#[test]
+fn transfer_closed_sender() {
+    let (main_tx, main_rx) = ipc::channel().unwrap();
+    let (transfer_tx, _) = ipc::channel::<()>().unwrap();
+    assert!(main_tx.send(transfer_tx).is_ok());
+    let _transferred_tx = main_rx.recv().unwrap();
+}
--- a/src/platform/macos/mod.rs
+++ b/src/platform/macos/mod.rs
@@ -537,7 +537,7 @@ impl OsOpaqueIpcChannel {
 }
 
 pub struct OsIpcReceiverSet {
-    port: Cell<mach_port_t>,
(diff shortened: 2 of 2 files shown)


src/test.rs
```diff
@@ -428,3 +428,11 @@ fn test_reentrant() {
     sender.send(null.clone()).unwrap();
     assert_eq!(null, receiver.recv().unwrap());
 }
+
+#[test]
+fn transfer_closed_sender() {
```
antrik: This is a bit nit-picky: but would you mind adding the test in a separate commit after the fix?

When making this sort of decisions, I like thinking in terms of both bisectability and revertability. Bisectability tells us that the test needs to go either in the same commit as the fix, or after it, since otherwise we would have a known-broken state in the history. Revertability says that it's better to have the test separate (either before or after the fix), since the test is probably valid regardless of the fix -- so if someone ever tries to back out the fix for any reason, the test should be still there and fail, thus making it clear that the issue needs to be addressed in some way... Considering both bisectability and revertability, we can conclude that a separate commit after the fix is indeed the best variant.

(Just to be clear, there are of course other situations as well: for tests that check newly added functionality, the same analysis would conclude that the tests need to go into the same commit as the change adding the functionality in question; and for tests that guard against regressions in things that should work even before the change, the test should go into a separate commit before the change...)

jdm:
I've addressed all of the comments. MachError::from can't easily handle the kernel errors because kernel_return_t and mach_msg_return_t are just aliases of the same type, and the range of errors that MachError::from handles overlaps with the range of the kernel return values.

antrik:
I know about the type aliasing -- but I don't understand why it is a problem? Isn't the whole idea behind this to allow uniform handling of all the errors?...

Doesn't really matter though, since I like the solution you implemented :-)

All the changes look very good to me. However, as I have limited experience with the `macos` back-end, I wonder whether it would be fine for me to r+ this myself; or better have someone else take another look?...

pcwalton:
Looks good to me.

Issue #214, opened 2018-12-12
Retrieve RawFd


oblique:
Is there any way that I can retrieve the underlying `RawFd`? I have a case that I want to set `O_CLOEXEC` flag and another case that I want to use `poll` to check for incoming data.

dlrobertson:
Yeah, we don't allow that at the moment. In theory we could implement `AsRawFd` on the OS specific backend if possible, but it would be **super** unsafe.

oblique:
Yes it would be unsafe in most of the cases. But what about marking it as unsafe and leave it to the developer?

dlrobertson:
Yeah, I have an implementation. I'll open a PR for it.

antrik:
I'm sceptical of this approach. A large part of the motivation for using a library like `ipc-channel` in the first place, is abstracting over different back-ends...

Regarding the specific use cases: for `O_CLOEXEC`, this sounds like something that would be useful to add through a proper interface across all back-ends I'd say?

As for `poll`, that's more tricky. What is your use case? Can it be covered with a different approach? Unless this is something very low-level and inherently platform-specific, polling over both IPC channels and different kinds of entities sounds like a use case for a some higher-level abstraction, such as features...

(Someone some initial work on features support for `ipc-channel` in the past, but it never went anywhere :-( )

oblique:
I wanted `poll` to check if there are data within some milliseconds. A different approach is to provide a `recv` that can timeout. `try_recv` is not good for me since it returns imminently.

antrik:
(Sorry for the late reply...)

This is a rather untypical use case... If there is no way to go about it differently, without needing a timeout in receive, I'd suggest using a Go-style approach: select over your data channel, along with a timer channel -- i.e. a dedicated channel with a sender that sends something once the timeout period passes.

(Or maybe just use a separate sender on the same channel to send the timeout message? In that case a simple `try_receive()` would suffice -- though it depends on the exact situation whether there is a way to make sure the timeout message isn't confused for real data...)

That's admittedly somewhat complicated -- but seems a much cleaner and more portable solution than some unsafe hack needing a dedicated interface...

PR #215, opened 2018-12-12
Implement AsRawFd if possible

dlrobertson:
## Summary

On unix-like OSes implement `AsRawFd` on `IpcReceiver`, `IpcSender`,
`IpcBytesReceiver`, and `IpcBytesSender`.

## Drawbacks

This introduces some OS specific code in `src/ipc.rs` and you can really
cause problems with a channel given the raw fd.

Fixes: #214

### DIFF
--- a/src/ipc.rs
+++ b/src/ipc.rs
@@ -19,6 +19,10 @@ use std::io::Error;
 use std::marker::PhantomData;
 use std::mem;
 use std::ops::Deref;
+#[cfg(all(not(feature = "force-inprocess"), any(target_os = "linux",
+                                                target_os = "openbsd",
+                                                target_os = "freebsd")))]
+use std::os::unix::io::{AsRawFd, RawFd};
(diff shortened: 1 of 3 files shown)


dlrobertson:
I don't really like this solution. I think there has got to be a better way. Closing this for now.

cgwalters:
I have a potential other use case for this.  Today the model encouraged by ipc-channel is to allocate a named socket in `/tmp` and have two processes arrange to communicate by sharing that path.

Yet, there's a much cleaner way in the case where there's a parent/child relationship, which is to just pass one half of a `socketpair()` from parent to child; e.g. on a well-known fd like `3`.

Yet as far as I can tell the relevant FD APIs to e.g. convert an `IpcSender` to/from a raw FD aren't exposed.

Oh also...this code should really be using `SOCK_CLOEXEC` for its fds.

PR #216, opened 2019-01-07
unix: Check libc not os, and zero msghdr private

### LINKED ISSUE #192: Doesn't compile with musl
I've tried to fix the compile issue on alpine/musl and ran into this problem:

`libc::msghdr` has two fields that don't exist on non-musl, `__pad1` and `__pad2`. Since there is no construcutor and these fields are private, `msghdr` can not be created.

https://github.com/rust-lang/libc/blob/master/src/unix/notbsd/linux/musl/b64/mod.rs

This currently blocks a musl port due to src/platform/unix/mod.rs. Everything else is trivial to fix.

makepost:
Fixes #192 similarly to what [audioipc-2](https://github.com/djg/audioipc-2/blob/1033cf7e4b5d14a138004bbb8708b2a1c3821939/audioipc/src/msg.rs#L79) does. Now builds on musl.

### DIFF
--- a/src/platform/unix/mod.rs
+++ b/src/platform/unix/mod.rs
@@ -42,14 +42,14 @@ const SCM_RIGHTS: c_int = 0x01;
 // Empirically, we have to deduct 32 bytes from that.
 const RESERVED_SIZE: usize = 32;
 
-#[cfg(target_os = "linux")]
+#[cfg(target_env = "gnu")]
 type IovLen = usize;
-#[cfg(target_os = "linux")]
+#[cfg(target_env = "gnu")]
 type MsgControlLen = size_t;
 
-#[cfg(any(target_os = "openbsd", target_os = "freebsd"))]
+#[cfg(not(target_env = "gnu"))]
 type IovLen = i32;
-#[cfg(any(target_os = "openbsd", target_os = "freebsd"))]
+#[cfg(not(target_env = "gnu"))]
 type MsgControlLen = socklen_t;
 
 unsafe fn new_sockaddr_un(path: *const c_char) -> (sockaddr_un, usize) {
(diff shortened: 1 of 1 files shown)


src/platform/unix/mod.rs
```diff
@@ -271,15 +271,14 @@ impl OsIpcSender {
...
-                    msg_flags: 0,
-                };
+                let mut msghdr: msghdr = unsafe { mem::zeroed() };
+                msghdr.msg_name = ptr::null_mut();
+                msghdr.msg_namelen = 0;
+                msghdr.msg_iov = iovec.as_mut_ptr();
+                msghdr.msg_iovlen = iovec.len() as IovLen;
+                msghdr.msg_control = cmsg_buffer as *mut c_void;
+                msghdr.msg_controllen = cmsg_space as MsgControlLen;
+                msghdr.msg_flags = 0;
```
dlrobertson: Could you wrap this into a function? E.g. `new_msghdr`?

makepost: Sure, re-uploaded.

makepost: Sorry for the noise, I seem to misunderstand how to extract this piece. Will get back to this.

src/platform/unix/mod.rs
```diff
@@ -999,15 +1003,7 @@ impl UnixCmsg {
...
-            msghdr: msghdr {
-                msg_name: ptr::null_mut(),
-                msg_namelen: 0,
-                msg_iov: iovec.as_mut_ptr(),
-                msg_iovlen: iovec.len() as IovLen,
-                msg_control: cmsg_buffer as *mut c_void,
-                msg_controllen: cmsg_length as MsgControlLen,
-                msg_flags: 0,
-            },
+            msghdr: new_msghdr(&mut iovec, cmsg_buffer, cmsg_length as MsgControlLen)
```
dlrobertson: `iovec` is already an `&mut [iovec]` no need for the `&mut`.

PR #218, opened 2019-01-12
ipc: implement Stream for the IpcBytesReceiver

dlrobertson:
If the async feature is enabled, implement the `Stream` trait for
`IpcBytesReceiver`.


asajeffrey:
LGTM. My only worry is that this like like it's wedded to futures 0.1, do you know what's likely to change for futures 0.3?

dlrobertson:
@asajeffrey any objections to me merging this in the meantime?

PR #227, opened 2019-03-04
Support futures 0.3 streams for IPC channels

### LINKED ISSUE #219: Update to futures 0.3
## Summary
Move from futures (0.1) to futures-preview (0.3).

## Details

Update the `futures` crate for `futures-preview` and update the implementation of `Stream` for `IpcBytesReceiver` and `IpcReceiver` to reflect this change. Between futures 0.1 and futures 0.3 the API was changed to include notification. In order to move to futures 0.3 `Stream::poll_next` needs to be able to use a `LocalWaker` to notify the task that we need to `poll` for new data.

asajeffrey:
Fixes #219


asajeffrey:
This is a breaking change to the API, so needs a bump to `0.12`.

The implementation uses `OpaqueIpcReceiver`, which isn't supported by `IpcBytesReceiver`, so I couldn't see how to implement `Stream` for it.

Since futures 0.3 is still only on nightly, this PR means that async is now only on nightly. We could try to maintain both, but that seems difficult.

asajeffrey:
Ping? @antrik

src/async.rs
```diff
@@ -0,0 +1,107 @@
...
+use std::task::Waker;
+use std::thread;
+
+/// A stream built from an IPC channel.
+pub struct IpcStream<T>(UnboundedReceiver<OpaqueIpcMessage>, PhantomData<T>);
+
+impl<T> Unpin for IpcStream<T> {}
+
+// A router which routes from an IPC channel to a stream.
+struct Router {
```
dlrobertson: `Router` and `ROUTER` are used in `router.rs`. Would it make sense to rename these to `AsyncRouter` and `ASYNC_ROUTER`?

asajeffrey: It's `AsyncRouter` vs `async::Router` which I don't have strong opinions about.

asajeffrey:
Review ping @antrik or @dlrobertson?

asajeffrey:
Argh, `async` is a keyword these days.
```
81 | pub mod async;
   |         ^^^^^ expected identifier, found reserved keyword
help: you can escape reserved keywords to use them as identifiers
   |
81 | pub mod r#async;
   |         ^^^^^^^
```

PR #234, opened 2019-07-08
fix: Build errors with async feature

Julusian:
This PR fixes a few build errors to make the async feature usable in 1.36 or nightly.

The `futures_api` feature is now stable in 1.36, so the flag can be removed.
And some other very minor typing fixes.


jdm:
These changes look ok to me, but it looks like we're missing a change to Cargo.toml to update the minimum version of the future crate to match the new API changes.

Julusian:
@jdm Rebased and the futures version is more specific now too

PR #236, opened 2019-07-17
Remove trait bounds on IpcSender and IpcReceiver.

### LINKED ISSUE #235: Remove Trait Bounds for Generic `T` in Receiver and Sender.
Currently IpcSender and Receiver are defined as:
```rust
pub struct IpcSender<T> where T: Serialize {
    ...
}
pub struct IpcReceiver<T> where T: for<'de> Deserialize<'de> + Serialize {
    ...
}
```

This makes it impossible to wrap either Ipc[Sender|Receiver] in my own type, and derive Deserialize and Serialize like so:
```rust
#[derive(Debug, Serialize, Deserialize)]
pub struct IpcReceiver<T> { // MY IpcReceiver
    receiver: ipc::IpcReceiver<(DetThreadId, T)>,
    ...
}
```
See https://github.com/serde-rs/serde/issues/1578 for details.

The solution is fairly straight forward.
1) Get rid of the trait bounds in the struct definition:
```diff
-pub struct IpcReceiver<T> where T: for<'de> Deserialize<'de> + Serialize {
+pub struct IpcReceiver<T> {

-pub struct IpcSender<T> where T: Serialize {
+pub struct IpcSender<T> {
```
2) Get rid of the trait bounds in the implementation blocks for Serialize and Deserialize:
```diff
-impl<'de, T> Deserialize<'de> for IpcReceiver<T> where T: for<'dde> Deserialize<'dde> + Serialize {
+impl<'de, T> Deserialize<'de> for IpcReceiver<T> {

-impl<T> Serialize for IpcReceiver<T> where T: for<'de> Deserialize<'de> + Serialize {
+impl<T> Serialize for IpcReceiver<T> {

-impl<'de, T> Deserialize<'de> for IpcSender<T> where T: Serialize {
+impl<'de, T> Deserialize<'de> for IpcSender<T> {

-impl<T> Serialize for IpcSender<T> where T: Serialize {
+impl<T> Serialize for IpcSender<T> {
```

Number 2 is interesting (I think?): the bounds on `T` are not required since the sender/receiver serialization implementation doesn't depend on `T` being serializable. So in general, it seems it is possible to serialize senders/receivers even when `T` itself is not serializable.

I can make this changes if there is no objection?

gatoWololo:
Remove trait bounds on IpcSender and IpcReceiver declaration.
Remove trait bounds on trait implemation for Serialize and Deserialize.

Fixes #235

### DIFF
--- a/src/ipc.rs
+++ b/src/ipc.rs
@@ -187,7 +187,7 @@ pub fn bytes_channel() -> Result<(IpcBytesSender, IpcBytesReceiver),Error> {
 ///
 /// [IpcReceiver]: struct.IpcReceiver.html
 #[derive(Debug)]
-pub struct IpcReceiver<T> where T: for<'de> Deserialize<'de> + Serialize {
+pub struct IpcReceiver<T> {
     os_receiver: OsIpcReceiver,
     phantom: PhantomData<T>,
 }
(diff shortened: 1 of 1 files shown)


emilio:
Makes sense to me, thanks.

@bors-servo r+

PR #246, opened 2019-11-07
Update crossbeam to 0.7 and crossbeam-channel to 0.4

mbrobbel:


nox:
(Note that this is a breaking change, we may want to lay down the upgrade path of ipc-channel in servo first.)

atouchet:
Note: There is another PR with similar changes as well #247

Eijebong:
Woops, hadn't seen this one... Mine doesn't break benchmarks though :p

nox:
Duplicate of #247.

PR #249, opened 2019-12-18
Make OpaqueIpcReceiver useful

Manishearth:
Currently it can't be converted to an IpcReceiver and isn't serializable, which makes it rather useless.

r? @pcwalton

### DIFF
--- a/src/ipc.rs
+++ b/src/ipc.rs
@@ -218,13 +218,7 @@ impl<T> IpcReceiver<T> where T: for<'de> Deserialize<'de> + Serialize {
 
 impl<'de, T> Deserialize<'de> for IpcReceiver<T> {
     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error> where D: Deserializer<'de> {
-        let index: usize = Deserialize::deserialize(deserializer)?;
-        let os_receiver =
-            OS_IPC_CHANNELS_FOR_DESERIALIZATION.with(|os_ipc_channels_for_deserialization| {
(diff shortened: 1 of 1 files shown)


pcwalton:
I'm somewhat uncomfortable reviewing this, because I don't recall the design constraints of `OpaqueIpcReceiver` anymore.

Manishearth:
cc @antrik

antrik:
I'm not sure I can help much with this: I never worked much with that part of of the code base. There may or may not have been a time where I fully understood the idea behind `OpaqueIpcSender` and `OpaqueIpcReceiver` -- but if there was, that's been a while ago...

As far as I can tell, the proposed change simply brings `OpaqueIpcReceiver` in line with `OpaqueIpcSender`? While that seems reasonable on the surface, I really can't tell whether there was some rationale why it hasn't been like that from the beginning...

What's the use case, anyway?

One thing I can say though is that `OpaqueIpcReceiver` is being used in `router.rs` -- so it's certainly not "useless" as is...

jdm:
#250 will need to merge before this one can.

oli-cosmian:
Could someone publish a new version with this change included?

jdm:
Opened https://github.com/servo/ipc-channel/pull/253 to start the process.

Manishearth:
This seems to already be included? https://docs.rs/ipc-channel/0.13.0/ipc_channel/ipc/struct.OpaqueIpcReceiver.html

oli-cosmian:
I just was confused. This PR was merged after `Cargo.toml` was last changed. Sorry for the noise

PR #256, opened 2020-01-31
macOS: Don't leak disconnected senders

jdm:


pcwalton:
Could you add a comment explaining why we use `mach_port_deallocate()` instead of `mach_port_mod_refs()`? This is subtle.

jdm:
The commit message includes the explanation; should I duplicate it in code comments?

pcwalton:
@jdm Probably a good idea, yeah. I'm worried someone will modify this code and reintroduce the bug.

PR #260, opened 2020-02-26
Add router shutdown

gterzian:
Adding a shutdown operation to the router, see for context https://github.com/servo/servo/pull/25685#pullrequestreview-364604068

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ipc-channel"
-version = "0.14.0"
+version = "0.14.1"
 description = "A multiprocess drop-in replacement for Rust channels"
 authors = ["The Servo Project Developers"]
 license = "MIT/Apache-2.0"
--- a/src/router.rs
+++ b/src/router.rs
@@ -52,6 +52,7 @@ impl RouterProxy {
             comm: Mutex::new(RouterProxyComm {
                 msg_sender: msg_sender,
                 wakeup_sender: wakeup_sender,
+                shutdown: false,
             }),
         }
     }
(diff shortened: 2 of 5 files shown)


gterzian:
@asajeffrey r?

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "ipc-channel"
-version = "0.14.0"
+version = "0.14.1"
```
CYBAI: The `RouterMsg` enum is added one more variant. I think it might be good to upgrade the minor version instead 👀?

gterzian: The enum is private, so I don't think it changes the api?

CYBAI: Ahh, I see; didn't notice that. sorry, then ignore my comment >_<

gterzian:
Oops, looks like a `cargo fmt` went through on the commit as well...

src/router.rs
```diff
@@ -53,19 +54,52 @@ impl RouterProxy {
                 msg_sender: msg_sender,
                 wakeup_sender: wakeup_sender,
             }),
+            shutdown: Mutex::new(false),
```
asajeffrey: Any time I see two locks, my deadlock alarms go off. Any chance you could make this an `AtomicBool` instead?

gterzian: Good point, and what I've done is put the boolean into the other locked data structure, what do you think(I find something like `compare_and_swap` really quite hard to work with)?

gterzian:
(I've also tried to make the test a bit clear, by only calling `shutdown` a second time at the end and adding a comment as to why).

gterzian:
@asajeffrey r? I've addressed your comment, although in a slightly different way from your initial suggestion. Please let me know what you think...

asajeffrey:
LGTM. @bors-servo r+

PR #264, opened 2020-03-10
Update mio to 0.7

nox:


atouchet:
Mio was updated to 0.8 in #326.

PR #270, opened 2020-12-26
Make it compile for wasm32-unknown-unknown and wasm32-wasi

tomaka:
This uses the "inprocess" implementation for the WebAssembly targets, like Windows, Android and iOS already do.

Note that this only compile with the master branch of the `uuid` crate, as https://github.com/uuid-rs/uuid/pull/477 hasn't been published yet.


emilio (approved):
Looks good, thanks.

PR #271, opened 2021-02-02
unix: always set FD_CLOEXEC on Linux

NeoRaider:
Always set FD_CLOEXEC to avoid child processes accidentally inheriting
file descriptors.

As none of the structs of ipc-channel allow extracting the underlying
file descriptors, there is no sane way to recover these FDs after an
exec anyways, so we don't need to make this configurable and just set
it unconditionally.

For now only Linux is covered, as special flags are available for all
relevant functions that set FD_CLOEXEC atomically, avoiding race
conditions in multithreaded processes:

- SOCK_CLOEXEC is passed to socket() and socketpair()
  - Supported since kernel 2.6.27
- MSG_CMSG_CLOEXEC is passed to recvmsg()
  - Supported since kernel 2.6.23
- MFD_CLOEXEC is passed to memfd_create()
- shmem_open() always sets FD_CLOEXEC according to POSIX

In libc, SOCK_CLOEXEC and MSG_CMSG_CLOEXEC are defined for FreeBSD and OpenBSD as well, so it's possible the restriction to `target_os = "linux"` is unnecessary after all, but I can't easily test on those OS.

This was previously discussed in #214.

### DIFF
--- a/src/platform/unix/mod.rs
+++ b/src/platform/unix/mod.rs
@@ -44,6 +44,17 @@ const SCM_RIGHTS: c_int = 0x01;
 // Empirically, we have to deduct 32 bytes from that.
 const RESERVED_SIZE: usize = 32;
 
+#[cfg(target_os = "linux")]
+const SOCK_FLAGS: c_int = libc::SOCK_CLOEXEC;
+#[cfg(not(target_os = "linux"))]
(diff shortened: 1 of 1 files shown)


jdm:
@bors-servo r+
Thanks!

PR #284, opened 2021-11-17
Try recv timeout

Xaeroxe:
There are currently two prominent ways to receive messages on the ipc channel, you can either block indefinitely, or not at all. For my use case, I'd like to have the lower CPU usage that comes with blocking, but also be able to exit the call periodically so my thread can check if it needs to exit. So I've added the `try_recv_timeout` method, and implemented it for all supported platforms. This method uses the platform facilities to block a recv call, but only for a set duration, at which point it will return `TryRecvError::Empty` if nothing came back.

### DIFF
--- a/src/platform/unix/mod.rs
+++ b/src/platform/unix/mod.rs
@@ -18,6 +18,7 @@ use libc::{setsockopt, size_t, sockaddr, sockaddr_un, socketpair, socklen_t, sa_
 use std::cell::Cell;
 use std::cmp;
 use std::collections::HashMap;
+use std::convert::TryInto;
 use std::error::Error as StdError;
 use std::ffi::CString;
 use std::fmt::{self, Debug, Formatter};
@@ -1070,7 +1071,11 @@ impl UnixCmsg {
             BlockingMode::Timeout(duration) => {
                 let events = libc::POLLIN | libc::POLLPRI | libc::POLLRDHUP;
                 let fd = &mut [libc::pollfd {fd, events, revents: 0}];
-                let result = libc::poll(fd.as_mut_ptr(), fd.len() as libc::c_ulong, duration.as_millis() as libc::c_int);
+                let result = libc::poll(
+                    fd.as_mut_ptr(),
+                    fd.len() as libc::c_ulong,
+                    duration.as_millis().try_into().unwrap_or(-1)
+                );
                 if result == 0 {
                     return Err(UnixError::Errno(EAGAIN));
                 } else if result < 0 {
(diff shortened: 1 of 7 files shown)


Xaeroxe:
This failure seems unrelated to my change

https://ci.appveyor.com/project/servo/ipc-channel/builds/41572812/job/iqg5fi9asp0exnec#L15

src/platform/unix/mod.rs
```diff
@@ -1054,10 +1060,23 @@ impl UnixCmsg {
...
-                return Err(UnixError::last())
+        match blocking_mode {
+            BlockingMode::Nonblocking => {
+                if libc::fcntl(fd, libc::F_SETFL, libc::O_NONBLOCK) < 0 {
+                    return Err(UnixError::last())
+                }
+            },
+            BlockingMode::Timeout(duration) => {
+                let events = libc::POLLIN | libc::POLLPRI | libc::POLLRDHUP;
+                let fd = &mut [libc::pollfd {fd, events, revents: 0}];
```
jdm, line 1073: Let's write this as `let mut fd = [...]` instead. I find that easier to reason about.

src/platform/unix/mod.rs
```diff
@@ -612,7 +612,8 @@ impl OsIpcOneShotServer {
             let socket_path = temp_dir.path().join("socket");
             let path_string = socket_path.to_str().unwrap();
 
-            let (sockaddr, len) = new_sockaddr_un(CString::new(path_string).unwrap().as_ptr());
+            let path_c_string = CString::new(path_string).unwrap();
+            let (sockaddr, len) = new_sockaddr_un(path_c_string.as_ptr());
```
jdm, line 617: Oh! Thank you for finding and fixing this.

jdm (approved):
This looks great! Thank you!

src/platform/windows/mod.rs
```diff
@@ -590,17 +593,22 @@ impl MessageReader {
             // issue the read to the buffer, at the current length offset
             self.r#async = Some(AliasedCell::new(AsyncData {
                 handle: self.handle.take(),
-                ov: NoDebug(Box::new(mem::zeroed())),
+                ov: NoDebug(Box::new({
+                    let mut overlapped: winapi::um::minwinbase::OVERLAPPED = mem::zeroed();
+                    // Create a manually reset event. The documentation for GetOverlappedResultEx
+                    // states you must do this in the remarks section.
+                    overlapped.hEvent = CreateEventA(ptr::null_mut(), winapi::shared::minwindef::TRUE, winapi::shared::minwindef::FALSE, ptr::null_mut());
```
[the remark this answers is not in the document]
Xaeroxe: Resolved in bec4ef6611490716eee3fdd52ce17707ff236836

src/platform/unix/mod.rs
Xaeroxe: Resolved in https://github.com/servo/ipc-channel/pull/284/commits/da99fcae7d890761881f8e0bbf035c4f995d357e

PR #294, opened 2022-03-29
chore: use `panic!("{}", ...)` in prep. for Rust 2021 edition

ErichDonGubler:
This PR paves the way for migration to Rust 2021 (if the team is willing). Even if Rust 2021 is not interesting for `ipc-channel`'s roadmap, it's one less lint for `rustc` to report!

### DIFF
--- a/src/platform/windows/aliased_cell.rs
+++ b/src/platform/windows/aliased_cell.rs
@@ -20,7 +20,7 @@ impl Drop for DropBomb {
         if thread::panicking() {
             eprintln!("{}", message);
         } else {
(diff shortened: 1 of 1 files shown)


ErichDonGubler:
Oh, derp! I didn't realize I'd filed an upstream PR. 😅 I've edited my PR OP to match what I intended to communicate here, sorry for the noise!

Let me know if the team here isn't interested in Rust 2021 for this crate -- I'm more than happy to amend the commit and PR OP to remove the expressed desire to transition to it if that's the case.

PR #295, opened 2022-05-03
Narrow the scope of unsafe blocks

peamaeq:
In this impl you use the unsafe keyword for many safe expressions. 

We need to mark unsafe operations more precisely using unsafe keyword. Keeping unsafe blocks small can bring many benefits. For example, when mistakes happen, we can locate any errors related to memory safety  within an unsafe block. This is the balance between Safe and Unsafe Rust. The separation is designed to make using Safe Rust as ergonomic as possible, but requires extra effort and care when writing Unsafe Rust. 

Hope this PR can help you.
Best regards.
**References**
https://doc.rust-lang.org/nomicon/safe-unsafe-meaning.html 
https://doc.rust-lang.org/book/ch19-01-unsafe-rust.html

### DIFF
--- a/src/platform/windows/mod.rs
+++ b/src/platform/windows/mod.rs
@@ -268,17 +268,15 @@ fn dup_handle_to_process_with_flags(handle: &WinHandle, other_process: &WinHandl
         return Ok(WinHandle::invalid());
     }
 
-    unsafe {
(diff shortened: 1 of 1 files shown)


jdm:
I don't think this is an improvement in maintainability.

PR #302, opened 2022-09-21
Port windows implementation from winapi to windows-rs

Xaeroxe:
`winapi` hasn't been updated since November of 2021, and the ecosystem seems to be making a general shift towards the officially supported `windows` crate instead. This PR moves `ipc-channel` to `windows`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,7 @@ description = "A multiprocess drop-in replacement for Rust channels"
 authors = ["The Servo Project Developers"]
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/servo/ipc-channel"
-edition = "2018"
+edition = "2021"
 
 [features]
 force-inprocess = []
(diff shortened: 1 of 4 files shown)


Xaeroxe:
Some deps are still using `winapi`, which makes this PR a bit silly.

https://github.com/XAMPPRocky/remove_dir_all/issues/43 that will eventually allow `tempfile` to remove the `winapi` dependency. They've already removed their own direct `winapi` dependency.

Xaeroxe:
This is ready to be merged now. `winapi` is no longer anywhere in the tree due to upstream changes.

Porges:
Any chance of getting this merged @jdm? It would also be good to upgrade `mio` to the 0.8 series given that there are other Rustsec advisories on some of its dependencies.

Porges:
@Xaeroxe I just tried the tests here and two of them are hanging:

- test::router_drops_callbacks_on_cloned_sender_shutdown
- test::router_drops_callbacks_on_sender_shutdown

Porges:
I've pushed a PR which further updates the `windows` crate to the current version, but the tests are still sticking there. Feel free to cherry-pick my change if you want.

Xaeroxe:
This was due to a subtle bug in the tests. The closure appeared to be capturing a value but in actuality was only capturing a reference to the value. I think rustc was auto-reffing the value. So what looked like an owning closure actually only owned a reference. I'm pushing a fix.

jdm (approved):

jdm:
Thank you for doing this work!

PR #313, opened 2023-06-05
Ensure macOS pointer addresses are aligned to eight bytes

jdm:
This is still a draft because it's not working yet.

### DIFF
--- a/src/platform/macos/mod.rs
+++ b/src/platform/macos/mod.rs
@@ -490,6 +490,7 @@ impl OsIpcSender {
         unsafe {
             let size = Message::size_of(&data, ports.len(), shared_memory_regions.len());
             let message = libc::malloc(size as size_t) as *mut Message;
+            println!("{:p}", message);
             (*message).header.msgh_bits = (MACH_MSG_TYPE_COPY_SEND as u32) |
                 MACH_MSGH_BITS_COMPLEX;
             (*message).header.msgh_size = size as u32;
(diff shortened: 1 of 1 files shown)


src/platform/macos/mod.rs
```diff
@@ -500,34 +501,38 @@ impl OsIpcSender {
             (*message).body.msgh_descriptor_count =
                 (ports.len() + shared_memory_regions.len()) as u32;
 
-            let mut port_descriptor_dest = message.offset(1) as *mut mach_msg_port_descriptor_t;
+            let mut port_descriptor_dest = (message as *mut PaddedMessage).offset(1) as *mut PaddedPortDescriptor;
```
mukilan, line 504: @jdm , I was able to fix the test failures with [these](https://github.com/mukilan/ipc-channel/commit/b58150a46d41f53c5ea2b79d9098f1cd4ef3ff4b) changes.

I'm not 100% sure about this, but this is my theory of what is causing the tests to fail: Because padding is added to the wrapped `mach_msg_port_descriptor_t` & `mach_ool_descriptor_t` structs, the entries in the array of descriptors following Message struct will have the wrong layout from the perspective of kernel. Since the kernel needs to interpret the values in these fields for validating the transfer of ports, it could end up reading random invalid data as the port names.

I couldn't find any documentation explaining the alignment requirements for OOL descriptor entries so I assumed they are byte-aligned (so address is now [u8; 8]). Relaxing the alignment requirement  allows the debug assertions inserted by rust on the reads/writes to succeed. The tests `platform::test::shared_memory` and `platform::test:shared_memory_clone` used to pass with Rust 1.67 and now continue to pass with my changes, so I think the Kernel doesn't have the 8-byte alignment requirement either.

I can raise a PR if the changes seem fine to you.

mukilan: Based on [this](https://github.com/apple/darwin-xnu/blob/main/osfmk/mach/message.h#L370) I think the kernel only expects 4-byte alignment. It also seems [rust-bindgen didn't translate pragma pack directives](https://github.com/rust-lang/rust-bindgen/issues/537) until 2018 and the [mach_sys.rs](https://github.com/servo/ipc-channel/commits/70123f70ac8baf1625d1ee9a2bec1aec6934a976/platform/macos/mach_sys.rs?browsing_rename_history=true&new_path=src/platform/macos/mach_sys.rs&original_branch=master) was added in 2015.

Should we just regenerate the mach_sys.rs using newer rust-bindgen? If so, I don't know which header file is the official one.

jdm: Oh, that makes a lot of sense! I'd say give rust-bindgen a try on the header that you linked and see what the output looks like?

jdm: Actually I guess https://github.com/apple-oss-distributions/xnu/blob/main/osfmk/mach/message.h is the official distribution now.

Issue #312, opened 2023-06-04
macOS builds are broken with recent rustc


jdm:
https://github.com/rust-lang/rust/pull/98112 identifies an unaligned write that occurs:
```
thread '<unnamed>' panicked at 'misaligned pointer dereference: address must be a multiple of 0x8 but is 0x7fa07c701059', src/platform/macos/mod.rs:539:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread '<unnamed>' panicked at 'panic in a function that cannot unwind', library/core/src/panicking.rs:126:5
stack backtrace:
   0:        0x104632e43 - std::backtrace_rs::backtrace::libunwind::trace::h8c19cce8ed8ccf51
                               at /rustc/398fa2187c88de46c13c142f600064483a563c86/library/std/src/../../backtrace/src/backtrace/libunwind.rs:93:5
   1:        0x104632e43 - std::backtrace_rs::backtrace::trace_unsynchronized::h991e04b43bc79443
                               at /rustc/398fa2187c88de46c13c142f600064483a563c86/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:        0x104632e43 - std::sys_common::backtrace::_print_fmt::h8987ba788c846a85
                               at /rustc/398fa2187c88de46c13c142f600064483a563c86/library/std/src/sys_common/backtrace.rs:65:5
   3:        0x104632e43 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h0a694d0e7599586c
                               at /rustc/398fa2187c88de46c13c142f600064483a563c86/library/std/src/sys_common/backtrace.rs:44:22
   4:        0x10464f49b - core::fmt::rt::Argument::fmt::h0a8aa1334c46e31c
                               at /rustc/398fa2187c88de46c13c142f600064483a563c86/library/core/src/fmt/rt.rs:138:9
   5:        0x10464f49b - core::fmt::write::h3efe3f464ccee79e
                               at /rustc/398fa2187c88de46c13c142f600064483a563c86/library/core/src/fmt/mod.rs:1094:21
   6:        0x104630148 - std::io::Write::write_fmt::h0e2bb4c22bdace59
                               at /rustc/398fa2187c88de46c13c142f600064483a563c86/library/std/src/io/mod.rs:1713:15
   7:        0x104632c3a - std::sys_common::backtrace::_print::h718192d6d9e6c0a5
[...]

PR #314, opened 2023-06-22
Regenerate mach bindings

mukilan:
The current 'mach_sys.rs' was generated before rust-bindgen started translating '#pragma pack(n)' directives into `#[repr(C, packed(n)]` attributes on the rust struct defintions. These missing attributes cause debug assertions to fail because of misaligned access to the fields when serializing and deserializing the structs for mach_msg calls (see #312 and [this](https://github.com/servo/ipc-channel/pull/313#discussion_r1230756392) discussion).

This PR regenerates the 'mach_sys.rs' file using the latest bindgen and mach headers from XCode MacOS SDK. It also introduces padding *before* the 'data size' field for messages with inline data so that they begin at 8-byte aligned addresses as required by rustc.

The second commit also fixes warnings emitted by latest rustc about drop() calls used on Copy types and drop() calls where a reference is passed instead of value.


jdm (approved):
Nice!

PR #320, opened 2023-08-23
Update to the latest version of rand

mrobinson:
This will allow updating the dependency in Servo and eliminating a
duplicated crate.


mrobinson:
@atouchet Thanks for the heads up. I will try to land that one.

Issue #222, opened 2019-01-17
Set up rustfmt


dlrobertson:
It would be great to find a rustfmt config that is close to the current style of the code along with a step in the CI that ensures code conforms to the style guide.

PR #323, opened 2023-08-28
Add rustfmt and clippy into CI

ikrivosheev:
Closes: https://github.com/servo/ipc-channel/issues/222

- [x] added rustfmt and clippy into CI
- [x] fixed clippy warnings 
- [x] replaced `actions-rs/toolchain` to `dtolnay/rust-toolchain`

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,14 +16,18 @@ jobs:
       matrix:
         features: ["unstable", "unstable force-inprocess", "unstable memfd"]
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - name: Install nightly toolchain
-        uses: actions-rs/toolchain@v1
+        uses: dtolnay/rust-toolchain@nightly
         with:
-          profile: minimal
-          toolchain: nightly
-          override: true
+          components: rustfmt, clippy
+
+      - name: rustfmt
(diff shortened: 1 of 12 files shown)


.github/workflows/main.yml
```diff
@@ -16,14 +16,18 @@ jobs:
...
-        uses: actions-rs/toolchain@v1
+        uses: dtolnay/rust-toolchain@nightly
         with:
-          profile: minimal
-          toolchain: nightly
-          override: true
+          components: rustfmt, clippy
+
+      - name: rustfmt
+        run: cargo fmt --features "${{ matrix.features }}"
```
mrobinson, line 27: This should probably run in `check` mode in order to fail on the CI if the input isn't formatted?

mrobinson (changes requested):
Thanks for this change! In order to verify the clippy fixes, would you be willing to split this PR into two? The first could do a rustfmt and then the second could fix the clippy issues. That would make this a bit easier to review. Thanks!

.github/workflows/main.yml
ikrivosheev: @mrobinson hello! thank you for the review, yes, i can.

ikrivosheev: First PR: https://github.com/servo/ipc-channel/pull/324. After merge it I will create new PR with clippy.

PR #280, opened 2021-08-23
Rust 2018 syntax and misc updates

lrazovic:
In this PR i wanted to take advantage of the idioms of Rust 2018, update the dependecies to the latest public version and remove deprecated code.

* `crossbeam-channel` updated from `0.4` to `0.5`.
* `rand` from `0.7` to `0.8`.
* `crossbeam-utils` from `0.7` to `0.8`.
* Removed the `extern crate` and `#[macro_use]` syntax, replaced by `use` where it's needed.
* Replaced `std::mem::replace` with `std::mem::take`.
* **[In `unsafe fn`]** Replaced the decprecated `mem::uninitialized` with `mem::MaybeUninit::uninit()`.
* **[In `unsafe` code block]** Replaced `ptr.offset(offset as isize)` with `ptr.add(offset)`. [relative clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#ptr_offset_with_cast)

Tested with:
* `cargo +nightly build --release --all-features --all-targets`
* `cargo +nightly test --all-features --all-targets`
* `cargo +nightly bench --all-features --all-targets`

I **didn't** formatted the code with `cargo fmt` because it would have brought too much changes, if necessary I'm ready to send a new patch.

PR #330, opened 2023-11-15
Remove usage of `extern crate`

mrobinson:
This is no longer necessary in recent versions of Rust.

This is a modified version of #280.


mrobinson:
Once the CI passes I'll merge this. I'm giving this an informal r+, which I cannot do officially since I opened this PR. I'm not really the original author of this change.

PR #335, opened 2024-05-04
Support empty IpcSharedMemory

### LINKED ISSUE #268: macos::OsIpcSharedMemory::from_bytes aborts if passed an empty slice

sagudev:
https://github.com/servo/ipc-channel/pull/334 made more simple.

Fixes #268.

### DIFF
--- a/src/ipc.rs
+++ b/src/ipc.rs
@@ -553,7 +553,8 @@ impl IpcReceiverSet {
     derive(PartialEq)
 )]
 pub struct IpcSharedMemory {
-    os_shared_memory: OsIpcSharedMemory,
+    /// None represents no data (empty slice)
+    os_shared_memory: Option<OsIpcSharedMemory>,
 }
 
 impl Deref for IpcSharedMemory {
@@ -561,7 +562,11 @@ impl Deref for IpcSharedMemory {
 
     #[inline]
     fn deref(&self) -> &[u8] {
-        &self.os_shared_memory
+        if let Some(os_shared_memory) = &self.os_shared_memory {
+            os_shared_memory
+        } else {
+            &[]
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/ipc.rs
```diff
@@ -589,32 +600,52 @@ impl Serialize for IpcSharedMemory {
...
     /// Create shared memory initialized with the bytes provided.
     pub fn from_bytes(bytes: &[u8]) -> IpcSharedMemory {
-        IpcSharedMemory {
-            os_shared_memory: OsIpcSharedMemory::from_bytes(bytes),
+        if bytes.is_empty() {
+            IpcSharedMemory::empty()
+        } else {
+            IpcSharedMemory {
+                os_shared_memory: Some(OsIpcSharedMemory::from_bytes(bytes)),
+            }
```
mrobinson, line 636: Could this be made even simpler by having this return `Option<IpcSharedMemory>`, much like `ptr::NonNull`? Then calling code would explicitly need to handle when the shared memory is empty. It feels like it should be an error to try to create an empty shared memory segment -- as it's kinda useless.

sagudev: I think that would made uses more verbose (most of the time there would be Some value so a lot of None values would be unreachable) and I feel this would complicate things for consumers. There is also no way in rust to tell that slice is at least length of 1, so consumers would still need to handle len=0 cases, even if they get Some. Also this would be breaking change and I tried to avoided that.

Empty buffer also has meaning (receiver was woken with empty message).

mrobinson: The thing is that this API needs to change anyway. Creating a shared memory segment can fail (see the manpage for `shm_open` for instance). ipc-channels is pretending like it cannot fail and it will panic if a failure does occur. When creating IPC channels, this reality is acknowledged, because `channel()` returns a `Result`. Not doing the same here feels like a bit of a lie. Many programs will want to handle failure gracefully.

sagudev: Indeed, but I think better error handling should be done as in separate PR. This PR just makes current API works as expected.

mrobinson (approved):

PR #341, opened 2024-07-24
Upgrade mio and release a new version of ipc-channels

mrobinson:


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "ipc-channel"
-version = "0.18.0"
+version = "0.19.0"
```
jdm, line 3: My read of the uses of mio in this codebase is that none of its types are exposed in ipc-channel's public API. We should be able to release 0.18.1 instead.

jdm (approved):

Cargo.toml
mrobinson: Makes sense to me! I've gone ahead and done that.

PR #342, opened 2024-07-24
Release version 0.18.2

mrobinson:
I previously tried to upgrade the version, but since I was using an old
checkout my change was simply eliminated from the commit during merge.
This PR actually cuts a new version (0.18.2).


mrobinson:
@jdm Apologies for the mixup. It seems I make a mistake in my previous PR.

PR #343, opened 2024-07-28
update recv to use correct values on FreeBSD

180watt:
It's rather unfortunate that Rust's libc doesn't know that FreeBSD has POLLRDHUP, but in the mean time we can add the value ourselves. We also need to use a uint for `nfds_t`. It seems that illumos also has POLLRDHUP, so this implementation could work there too. (Testing welcome.) I'm a novice when it comes to Rust, so any suggestions to this patch is very welcomed.

P.S. I failed to mention this, but in its current state (pre-patch), ipc-channel fails to build on FreeBSD

### DIFF
--- a/src/platform/unix/mod.rs
+++ b/src/platform/unix/mod.rs
@@ -1162,7 +1162,18 @@ impl UnixCmsg {
                 }
             },
             BlockingMode::Timeout(duration) => {
-                let events = libc::POLLIN | libc::POLLPRI | libc::POLLRDHUP;
+                #[cfg(target_os = "linux")]
+                const POLLRDHUP: lib::c_short = libc::POLLRDHUP;
+
+                // It's rather unfortunate that Rust's libc doesn't know that
+                // FreeBSD has POLLRDHUP, but in the mean time we can add
+                // the value ourselves.
+                // https://cgit.freebsd.org/src/tree/sys/sys/poll.h#n72
+                #[cfg(target_os = "freebsd")]
+                const POLLRDHUP: libc::c_short = 0x4000;
+
+                let events = libc::POLLIN | libc::POLLPRI | POLLRDHUP;
+
                 let mut fd = [libc::pollfd {
                     fd,
                     events,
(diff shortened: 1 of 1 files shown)


wusyong:
Can you either change only the parts that are affected, like `POLLRDHUP` and `poll,` or add a comment for the method to describe the difference between Linux and FreeBSD?

180watt:
Updated my commit to be more "surgical." Thanks for the suggestions.

src/platform/unix/mod.rs
```diff
@@ -1162,18 +1162,36 @@ impl UnixCmsg {
...
+                #[cfg(target_os = "linux")]
                 let events = libc::POLLIN | libc::POLLPRI | libc::POLLRDHUP;
+
+                // It's rather unfortunate that Rust's libc doesn't know that
+                // FreeBSD has POLLRDHUP, but in the mean time we can add
+                // the value ourselves.
+                // https://cgit.freebsd.org/src/tree/sys/sys/poll.h#n72
+                #[cfg(target_os = "freebsd")]
+                let events = libc::POLLIN | libc::POLLPRI | 0x4000;
+
```
jdm, line 1174: Could we instead add a POLLRDHUP constant that is conditionally defined, rather than the implementation code here?

180watt: Most certainly.

wusyong (approved):

src/platform/unix/mod.rs
```diff
@@ -1162,18 +1162,36 @@ impl UnixCmsg {
...
                     fd.as_mut_ptr(),
                     fd.len() as libc::c_ulong,
                     duration.as_millis().try_into().unwrap_or(-1),
                 );
 
+		// FreeBSD's nfds_t is an unsigned integer.
+		#[cfg(target_os = "freebsd")]
+                let result = libc::poll(
+                    fd.as_mut_ptr(),
+                    fd.len() as libc::c_uint,
```
jdm, line 1191: We should be able to use `as _` and avoid any platform-specific code.

180watt: Fixed up.

180watt:
If this patch can also cause a bump in minor version, that would be greatly appreciated.

jdm (approved):
Thanks!

180watt:
I have missed a critical `c` in the code, and we are now trying to do `lib::c_short`. Of course this had to be in the path that I _don't_ build, so I didn't see the error. I am **very** sorry about that. I am also currently away from my computer and can not readily push something. Do you mind either reverting or pushing a fix?

PR #346, opened 2024-07-29
Update windows-rs to 0.58

### LINKED ISSUE #345: Update windows-rs
Latest version is 0.58, we use 0.48.

sagudev:
Fix #345.

My first time working with windows-rs, so review carefully. I tried to stick to the current code schematic, but the whole codebase  would need a cleanup (a lot of clippy warnings).

Draft because I want a companion PR in servo.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ crossbeam-utils = "0.8"
 static_assertions = "1.1.0"
 
 [target.'cfg(target_os = "windows")'.dependencies.windows]
-version = "0.48.0"
+version = "0.58.0"
 features = [
     "Win32_Foundation",
     "Win32_System_WindowsProgramming",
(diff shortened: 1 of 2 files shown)


atouchet:
Servo now has windows 0.58 in its dependencies. Does this still need to remain a draft?

sagudev:
> Servo now has windows 0.58 in its dependencies. Does this still need to remain a draft?

Not really, I just forgot.

mrobinson (approved):
I am by no means an expert in the Win32 API, but looked over this a couple times and I don't see any obvious issues. If it works, I say let's go for it.

sagudev:
> I am by no means an expert in the Win32 API, but looked over this a couple times and I don't see any obvious issues. If it works, I say let's go for it.

The problem is that even that if it does not works the bugs are subtile (most changes are in error handling which happens rarely, and we only run smoketest for win, we need windows-wpt).

Issue #126, opened 2016-11-28
Zero-copy support


siriux:
Is it possible to do zero-copy with ipc-channel today? I've seen OsIpcSharedMemory, but I'm not sure it can be used directly for this purpose.

If it's not possible right now, do you plan to support it? I'm testing some ideas for interprocess video transfer, and zero-copy would be really helpful.

Thanks

emilio:
I think this was discussed, and the conclusion was something like: Even though it's hard to expose a safe API on top of shmem to avoid races, letting the user write to a shared memory buffer with an unsafe API could be desirable, but it's just not done.

It should be a straight-forward addition I think, a patch would probably be welcomed :)

cc @antrik, did I get that right?

siriux:
I'm quite new to Rust, but I might try to write a patch when I get more confidence. Thanks

antrik:
We need to clarify the use case here.

What `IpcSharedMemory` currently allows, is creating a shared memory object initialised with some pre-existing data, and sending that to other processes, where it is mapped, so the data can be read there. This is *not* zero-copy, since creation of the object allocates new memory that is initialised by copying some existing data.

This could be fixed by extending the API to allow decoupling the allocation of the buffer from initialisation. This way we could create a new buffer for each frame, then fill it with data, and finally send it to the consumer process, where it is read and finally freed.

One possible approach for that would be introducing some kind of proto-SHM-object, which is allocated but not initialised, and can be borrowed mutably to fill it with data, before it's converted into a regular immutable SHM object for sending. Another approach would be allowing some kind of callback for supplying the initialisation data -- this would be less flexible in how the data can be supplied (probably precluding DMA?), but keep the API entirely safe, by not exposing uninitialised memory to users.

An entirely different case would arise if we want permanent shared memory, that can be overwritten by the producer process with new data repeatedly even after sending it off to a consumer process. This would be much more involved, since it requires some kind of inter-process locking to make it safe. Wrapping this in some kind of safe abstraction might be somewhat tricky I suspect, though I haven't spent too much thought on it. (And I'm not really familiar with inter-process locking mechanisms.)

siriux:
Thanks Olaf.

What I had in mind was more or less what you described, borrowing mutably
to initialize directly on the shared buffer, as it's more flexible. But if
you want to be fast, I think some kind of reusing is needed too.

In my opinion, the easiest way would be to keep the same simple interface
(+ mutable borrow), where every time you borrow a new mutable buffer it can
[...]

PR #356, opened 2024-08-20
feat: Allow reusing same shared IpcSharedMem for transfers

sagudev:
Simple "solution" to https://github.com/servo/ipc-channel/issues/126, it's not exactly zero-copy, but it allows to modify inner buffer of IpcSharedMem and then send it back. This is all `unsafe` but we can uphold all safety variants for `GPUBuffer` (request-response msgs).

Example

before:
```rust
const SIZE: usize = 50;
let (tx1, rx1) = ipc::channel().unwrap();
let (tx2, rx2) = ipc::channel().unwrap();
std::thread::spawn(move || {
    let mut data = rx1.recv().unwrap() // allocation 1
                                      .to_vec(); // allocation 2
    for i in 0..=SIZE {
        *data[i] = i as u8;
    }
    tx2.send(IpcSharedMemory::from_bytes(&data)).unwrap(); // allocation 3
});
tx1.send(IpcSharedMemory::from_byte(0, SIZE)).unwrap();
rx2.recv().unwrap();
```
after:
```rust
const SIZE: usize = 50;
let (tx1, rx1) = ipc::channel().unwrap();
let (tx2, rx2) = ipc::channel().unwrap();
std::thread::spawn(move || {
    // SAFETY: This is safe because IpcSharedMemory is constructed in place from main thread (will only be accessed by send)
    let mut ism = unsafe{ rx1.recv().unwrap() }; // allocation 1
    {
        let data = ism.deref_mut();
        for i in 0..=SIZE {
            *data[i] = i as u8;
        }
    }
    tx2.send(ism).unwrap();
});
tx1.send(IpcSharedMemory::from_byte(0, SIZE)).unwrap();
rx2.recv().unwrap();
```

### DIFF
--- a/src/ipc.rs
+++ b/src/ipc.rs
@@ -545,6 +545,24 @@ impl Deref for IpcSharedMemory {
     }
 }
 
+impl IpcSharedMemory {
+    /// Returns a mutable reference to the deref of this [`IpcSharedMemory`].
+    ///
+    /// # Safety
+    ///
+    /// This is safe if there is only one reader/writer on the data.
+    /// User can achieve this by not cloning [`IpcSharedMemory`]
+    /// and serializing/deserializing only once.
+    #[inline]
+    pub unsafe fn deref_mut(&mut self) -> &mut [u8] {
+        if let Some(os_shared_memory) = &mut self.os_shared_memory {
+            os_shared_memory.deref_mut()
+        } else {
+            &mut []
+        }
+    }
+}
+
 impl<'de> Deserialize<'de> for IpcSharedMemory {
     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
     where
--- a/src/platform/unix/mod.rs
+++ b/src/platform/unix/mod.rs
@@ -32,7 +32,7 @@ use std::hash::BuildHasherDefault;
 use std::io;
 use std::marker::PhantomData;
 use std::mem;
-use std::ops::{Deref, RangeFrom};
+use std::ops::{Deref, DerefMut, RangeFrom};
 use std::os::fd::RawFd;
 use std::ptr;
 use std::slice;
@@ -864,6 +864,13 @@ impl Deref for OsIpcSharedMemory {
     #[inline]
     fn deref(&self) -> &[u8] {
         unsafe { slice::from_raw_parts(self.ptr, self.length) }
+    }
+}
+
+impl OsIpcSharedMemory {
+    #[inline]
+    pub unsafe fn deref_mut(&mut self) -> &mut [u8] {
+        unsafe { slice::from_raw_parts_mut(self.ptr, self.length) }
     }
 }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,10 @@ harness = false
 name = "ipc_receiver_set"
 harness = false
 
+[[bench]]
+name = "ipc_shared_mem"
+harness = false
+
 [features]
 default = []
 force-inprocess = []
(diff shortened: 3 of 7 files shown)


sagudev:
Benches from CI: https://github.com/servo/ipc-channel/actions/runs/10472511882

src/ipc.rs
```diff
@@ -570,6 +570,24 @@ impl Deref for IpcSharedMemory {
...
+impl IpcSharedMemory {
+    /// Returns a mutable reference to the deref of this [`IpcSharedMemory`].
+    ///
+    /// # Safety
+    ///
+    /// This is safe if there is only one reader/writer on the data.
+    /// User can achieve this by not cloning [`IpcSharedMemory`]
+    /// and serializing/deserializing only once.
+    #[inline]
+    pub unsafe fn deref_mut(&mut self) -> &mut [u8] {
```
jschwe, line 582: I'm a bit confused on how this works. Do we know that if this code is running, then the other process is guaranteed to not be be able to access this memory?

src/platform/unix/mod.rs
```diff
@@ -866,6 +866,13 @@ impl Deref for OsIpcSharedMemory {
     }
 }
 
+impl OsIpcSharedMemory {
+    #[inline]
+    pub unsafe fn deref_mut(&mut self) -> &mut [u8] {
```
jschwe, line 871: In general is `implementing`  deref / deref_mut sound for shared memory? Since Rust / LLVM doesn't see the other process accessing the memory, wouldn't we be required to use volatile reads and writes to prevent unwanted optimizations?

src/ipc.rs
sagudev: no there is no such guarantee, hence the method is marked unsafe (we will probably need named locks and serialization that takes ownership)

src/platform/unix/mod.rs
sagudev: no, hence marked as unsafe, same safety docs also applies to this.

src/ipc.rs
sagudev: users needs to make sure that only one process/thread uses memory (no synchronization is guaranteed as user must send ipcsharedmem back), this works if this shared memory is used for something like req-resp, for example webgpu's mapasync where we map (clone) buffer to JS thread (while it's mapped GPU will not touch it) then when unmaping we send this data back.

src/platform/unix/mod.rs
jschwe: This question is slightly different - let me rephrase it:
Lets assume the user upholds all Safety invariants - Is it possible that LLVM still breaks our code, because it e.g. sees that the written values are never read (in the current process) and thus optimizes out some writes (or reads)

sagudev: I think not, because they are still read when sending back.

src/ipc.rs
sagudev: I update PR description with example of usage (there is also bench in PR that can act as example).

jdm (approved):
This seems reasonable as a fully-unsafe API. It would be nice if we could detect invalid usage at runtime similar to Arc::get_mut, but I don't have ideas for how to do that soundly.

sagudev:
> This seems reasonable as a fully-unsafe API. It would be nice if we could detect invalid usage at runtime similar to Arc::get_mut, but I don't have ideas for how to do that soundly.

The main problem is we would need to do cross-process locking which is costly, but we might want to implement this in the future to also fulfill other usecases.

PR #359, opened 2024-09-19
Fix user after free in MacOS select()

webbeef:
Update to the loop handling large messages logic in order to not reuse `message` after it's backing buffer was deallocated.

cc @jdm since you wrote this in the first place.


jdm (approved):
Oof. Thanks for tracking this down!

jdm:
Published 0.18.3.

PR #362, opened 2024-10-11
Expose `OpaqueIpcMessage` as `IpcMessage`

mrobinson:
This makes the API a bit friendlier and future-proof. Instead of
returning a tuple from raw `recv()` return the `IpcMessage` struct that
is used internally in the crate. In addition, this is used to pass data
from the platform layer -- removing clippy warnings about complex data
types. One upside to this is that `Option<IpcSharedMemory>` is turned
into `IpcSharedMemory` in some places where having a `None` value
shouldn't be possible.

This simplification removes about 300 lines of code.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ name = "ipc_receiver_set"
 harness = false
 
 [features]
+default = []
 force-inprocess = []
 memfd = ["sc"]
 async = ["futures", "futures-test"]
--- a/src/ipc.rs
+++ b/src/ipc.rs
@@ -251,16 +251,13 @@ where
 {
     /// Blocking receive.
     pub fn recv(&self) -> Result<T, IpcError> {
-        let (data, os_ipc_channels, os_ipc_shared_memory_regions) = self.os_receiver.recv()?;
-        OpaqueIpcMessage::new(data, os_ipc_channels, os_ipc_shared_memory_regions)
-            .to()
-            .map_err(IpcError::Bincode)
+        self.os_receiver.recv()?.to().map_err(IpcError::Bincode)
     }
 
     /// Non-blocking receive
     pub fn try_recv(&self) -> Result<T, TryRecvError> {
-        let (data, os_ipc_channels, os_ipc_shared_memory_regions) = self.os_receiver.try_recv()?;
-        OpaqueIpcMessage::new(data, os_ipc_channels, os_ipc_shared_memory_regions)
+        self.os_receiver
+            .try_recv()?
             .to()
             .map_err(IpcError::Bincode)
             .map_err(TryRecvError::IpcError)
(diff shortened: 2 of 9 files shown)


mrobinson:
This will need #364 to work properly.

Cargo.toml
```diff
@@ -20,6 +20,7 @@ name = "ipc_receiver_set"
 harness = false
 
 [features]
+default = ["async"]
```
jdm, line 23: Why this change?

src/ipc.rs
```diff
@@ -735,14 +724,17 @@ impl OpaqueIpcMessage {
...
+                            .into_iter()
+                            .map(Some)
+                            .collect(),
                     );
                     let result = bincode::deserialize(&self.data[..]);
-                    mem::swap(
+                    let _ = mem::replace(
                         &mut *os_ipc_shared_memory_regions_for_deserialization.borrow_mut(),
-                        &mut self.os_ipc_shared_memory_regions,
+                        old_ipc_shared_memory_regions_for_deserialization,
```
jdm, line 737: I think this would be equivalent to (and more clear as):
```rust
*os_ipc_shared_memory_regions_for_deserialization.borrow_mut() = old_ipc_shared_memory_regions_for_deserialization;
```

jdm (approved):
This is a really nice reduction in complexity!

src/ipc.rs
mrobinson: :+1:

mrobinson:
Thanks for the review!

src/ipc.rs
```diff
@@ -699,19 +691,16 @@ impl Debug for OpaqueIpcMessage {
     }
 }
 
-impl OpaqueIpcMessage {
-    fn new(
+impl IpcMessage {
+    pub(crate) fn new(
```
mukilan, line 695: I'm curious if there is reason for not marking this as `pub`. The struct only has pub fields, so it doesn't seem like we are gaining much in terms of encapsulation.

mrobinson: Fair enough. I think mainly this wasn't marked as `pub` because it isn't expected that a user of `ipc-channel` would ever need to create this data structure. `pub(crate)` is useful in this case because the compiler can determine if the function ever becomes unused.

mukilan: > because it isn't expected that a user of ipc-channel would ever need to create this data structure

Should we then remove the 'pub' from the fields? Having them public means it will become part of our API in the next release and OpaqueIpcMessage only [had private fields](https://docs.rs/ipc-channel/latest/ipc_channel/ipc/struct.OpaqueIpcMessage.html). I'm happy to do the change.

mrobinson: I am not opposed to making the fields private again.

PR #364, opened 2024-10-11
Remove `windows-shared-memory-equality` feature

mrobinson:
This feature is useful because it allows compiling for versions of
Windows older than Windows 10. Windows 10 was released about a decade
ago, so I think it's fine to remove support for older versions of
Windows. For example Windows 7 is no longer supported as of January 2020
and it seems that less than 3% of users are using it.


jdm (approved):
Agreed!

Issue #239, opened 2019-07-26
Able to convert `OpaqueIpcMessage` to "smaller" sized types.


gatoWololo:
The following minimal example reproduces the error (I'm using the latest crates.io version of `ipc-channel` but I bet it works with latest commit on master as well):
```rust
use ipc_channel::ipc;
use ipc_channel::router::ROUTER;

fn main() -> Result<(), std::io::Error> {
    let (sender, receiver) = ipc::channel::<(i32, i32)>()?;
    // Use this second channel to force main thread to block until child thread
    // is done.
    let (sender2, receiver2) = ipc::channel()?;

    ROUTER.add_route(receiver.to_opaque(), Box::new(move |msg|{
        match msg.to::<i32>() {
            Ok(val) => println!("Val: {:?}", val),
            Err(e) =>  println!("Error: {:?}", e),
        };
        sender2.send(());
    }));

    sender.send((1234, 5678)).unwrap();
    receiver2.recv().unwrap();
    Ok(())
}
```
Which happily outputs: `Val: 1234`.

Basically I'm able to convert the `(i32, i32)` to a `i32` through `to::<i32>()`. I only tried it for numerical types, but it seems we have observed this behavior for less trivial types: https://github.com/servo/servo/issues/23818#issuecomment-514877686

I'm planning to (ab)use this behavior... but it seems a little undefined-behaviory to me?

PR #365, opened 2024-10-14
Implement typed routes for `ROUTER`

### LINKED ISSUE #242: Add Typed Version of `add_route`.
Currently `add_route` is defined as:
```rust
pub fn add_route(&self, receiver: OpaqueIpcReceiver, callback: RouterHandler);
pub type RouterHandler = Box<FnMut(OpaqueIpcMessage) + Send>;
```
It is up to the user to remember and properly turn the OpaqueIpcMessage into the correct type. This leads to wrong types being used as in https://github.com/servo/servo/issues/23818

Instead I recommend adding a typed version as:
```rust
pub fn add_route<T>(&self, receiver: IpcReceiver<T>, mut callback: RouterHandler<T>)
```
The user no longer has to learn about `OpaqueIpcReceiver` and it is compile-time checked, so the type can never be used wrong! Internally, the implementation converts the `OpaqueIpcMessage` into `T` so it can never be misused:
```rust
|msg: OpaqueIpcMessage| callback(msg.to::<T>().unwrap());
```

Eventually the existing `add_route` could be deprecated in favor of the typed one.

simonwuelker:
Implements `ROUTER::add_typed_route`, which (unlike `ROUTER::add_route`) knows about the type expected by the callback and makes it impossible to mismatch types between the receiver and the callback.

This method can be used pretty much exactly like `ROUTER::add_route`, I added a small test case that demonstrates this.

Closes #242

### DIFF
--- a/src/router.rs
+++ b/src/router.rs
@@ -12,6 +12,7 @@
 //! `IpcReceiver<T>`s. The router will then either call the appropriate callback or route the
 //! message to a crossbeam `Sender<T>` or `Receiver<T>`. You should use the global `ROUTER` to
 //! access the `RouterProxy` methods (via `ROUTER`'s `Deref` for `RouterProxy`.
+
 use lazy_static::lazy_static;
 use std::collections::HashMap;
 use std::sync::Mutex;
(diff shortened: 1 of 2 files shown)


jdm (approved):
This is a great usability improvement! Thanks.

mrobinson:
> This is a great usability improvement! Thanks.

Agreed! Unfortunately, `add_route` is a bit easy to use because of #239. Does it make sense to mark it as `unsafe`?

simonwuelker:
> Agreed! Unfortunately, add_route is a bit easy to use because of https://github.com/servo/ipc-channel/issues/239. Does it make sense to mark it as unsafe?

`unsafe` is usually reserved for functions that can cause undefined behavior if used incorrectly, and I don't think `add_route` can quite do that.h

Perhaps a
`#[deprecated(since="0.18.4", note="please use 'add_typed_route' instead")]`
annotation would be more appropriate?

simonwuelker:
Hmm, perhaps I should have implemented the callback to receive a `Result<T, Error>` instead.
Functions like servo's [`route_ipc_receiver_to_new_crossbeam_receiver_preserving_errors`](https://github.com/servo/servo/blob/c37fb2e45300a58e97fcc7d2f00ecf7db1f0fc72/components/constellation/constellation.rs#L591) need to know about any errors that occured, because constellation is not allowed to crash.

I'll make a follow-up PR once we have decided whether to mark `add_route` as `unsafe`/`deprecated`.

jdm:
Agreed that deprecated feels more appropriate here than unsafe. If there isn't a way to trigger memory unsafety, unsafe would be overkill.

mrobinson:
I guess since serializable types cannot contain pointers, it is not possible to create memory unsafety with this issue. 🤔

PR #367, opened 2024-10-15
Mark `Router::add_route` as deprecated

simonwuelker:
* Marks `Router::add_route` as deprecated as discussed in https://github.com/servo/ipc-channel/pull/365#issuecomment-2413469573
* Creates a new minor release (that someone with the relevant permissions should push to crates.io). This is necessary for the `#[deprecated]` annotation
* Makes the callback of `Router::add_typed_route` receive a `Result<T, bincode::Error>` (https://github.com/servo/ipc-channel/pull/365#issuecomment-2413540813)

### DIFF
--- a/src/router.rs
+++ b/src/router.rs
@@ -83,7 +83,7 @@ impl RouterProxy {
     {
         // Before passing the message on to the callback, turn it into the appropriate type
         let modified_callback = move |msg: IpcMessage| {
-            let typed_message = msg.to::<T>().unwrap();
+            let typed_message = msg.to::<T>();
             callback(typed_message)
         };
         self.add_route(receiver.to_opaque(), Box::new(modified_callback));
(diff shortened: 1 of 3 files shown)


src/router.rs
```diff
@@ -59,6 +59,10 @@ impl RouterProxy {
 
     /// Add a new (receiver, callback) pair to the router, and send a wakeup message
     /// to the router.
+    ///
+    /// Consider using [add_typed_route](Self::add_typed_route) instead, which prevents
+    /// mismatches between the receiver and callback types.
+    #[deprecated(since = "0.19", note = "please use 'add_typed_route' instead")]
```
jdm, line 65:
```suggestion
    #[deprecated(since = "0.19.0", note = "please use 'add_typed_route' instead")]
```

jdm (approved):

PR #368, opened 2024-10-18
Use more constants and functions from libc

sunshowers:
I've been trying to port ipc-channel to illumos for use in a downstream project, and I found that this code was using a local copy of several functions and constants that are in libc. In particular:

* Some of the `CMSG_` functions were incorrect on illumos.
* `SCM_RIGHTS` was also not correct.

The correct versions of these functions are in libc, so use them. (I had to bump libc to 0.2.161 to pull in the definition of `POLLRDHUP` on FreeBSD.)

This is an intermediate commit that should hopefully be uncontroversial. The full test suite still doesn't pass on illumos yet, but I'm still working on it (in my spare time).

### DIFF
--- a/src/platform/unix/mod.rs
+++ b/src/platform/unix/mod.rs
@@ -11,21 +11,22 @@ use crate::ipc::{self, IpcMessage};
 use bincode;
 use fnv::FnvHasher;
 use lazy_static::lazy_static;
-use libc::{self, MAP_FAILED, MAP_SHARED, PROT_READ, PROT_WRITE, SOCK_SEQPACKET, SOL_SOCKET};
+use libc::{
+    self, cmsghdr, linger, CMSG_DATA, CMSG_LEN, CMSG_SPACE, MAP_FAILED, MAP_SHARED, PROT_READ,
+    PROT_WRITE, SOCK_SEQPACKET, SOL_SOCKET,
+};
 use libc::{c_char, c_int, c_void, getsockopt, SO_LINGER, S_IFMT, S_IFSOCK};
-use libc::{iovec, mode_t, msghdr, off_t, recvmsg, sendmsg};
+use libc::{iovec, msghdr, off_t, recvmsg, sendmsg};
 use libc::{sa_family_t, setsockopt, size_t, sockaddr, sockaddr_un, socketpair, socklen_t};
 use libc::{EAGAIN, EWOULDBLOCK};
 use mio::unix::SourceFd;
 use mio::{Events, Interest, Poll, Token};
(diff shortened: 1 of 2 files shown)


sunshowers:
Oh whoops, forgot to use libc's memfd :)

src/platform/unix/mod.rs
```diff
@@ -1096,7 +1099,7 @@ fn create_shmem(name: CString, length: usize) -> c_int {
 #[cfg(all(feature = "memfd", target_os = "linux"))]
 fn create_shmem(name: CString, length: usize) -> c_int {
     unsafe {
-        let fd = memfd_create(name.as_ptr(), libc::MFD_CLOEXEC as usize);
+        let fd = libc::memfd_create(name.as_ptr(), libc::MFD_CLOEXEC);
```
sunshowers, line 1102: Oh hmm, glibc support was only added in 2.27. Not sure what the glibc policy of ipc-channel is.

sunshowers: FWIW I think the old call was incorrect, since it passed in a usize but the syscall is a cint: https://github.com/torvalds/linux/blob/4d939780b70592e0f4bc6c397e52e518f8fb7916/mm/memfd.c#L332

sunshowers: Looks like glibc 2.27 came out in 2018: https://sourceware.org/glibc/wiki/Release/2.27.

src/platform/unix/mod.rs
```diff
@@ -991,7 +991,10 @@ fn recv(fd: c_int, blocking_mode: BlockingMode) -> Result<IpcMessage, UnixError>
         let channel_length = if cmsg_length == 0 {
             0
         } else {
-            (cmsg.cmsg_len() - CMSG_ALIGN(mem::size_of::<cmsghdr>())) / mem::size_of::<c_int>()
+            // The control header is followed by an array of FDs. The size of the control header is
+            // determined by CMSG_SPACE. (On Linux this would the same as CMSG_ALIGN, but that isn't
+            // exposed by libc. CMSG_SPACE(0) is the portable version of that.)
+            (cmsg.cmsg_len() - CMSG_SPACE(0) as size_t) / mem::size_of::<c_int>()
```
sagudev, line 997: If `CMSG_ALIGN` is not in libc we should add it.

sunshowers:
Hey -- how can I help get this landed? I can revert the glibc change if 2.27 is too new, though note that it covers Ubuntu 18.04, CentOS 8 and Debian buster.

jdm (approved):
This seems reasonable.

Issue #101, opened 2016-09-14
ipc-channel makes assumptions about OS buffer sizes


vvuk:
ipc-channel seems to make assumptions about underlying OS buffer sizes; the tests want to send()/recv() in the same thread on the same channel, e.g.: https://github.com/servo/ipc-channel/blob/master/src/platform/test.rs#L97 .  This is a problem; is this intended in the API, or is it unintentional due to the way the tests are written?

I'm going to end up needing to queue writes to happen from another thread for every sender as a result of this for the windows impl (probably a work queue).

nox:
I'm pretty sure you are supposed to be able to send and recv on the same thread.

vvuk:
The `big_data` test at https://github.com/servo/ipc-channel/blob/master/src/platform/test.rs#L128 explicitly makes a new thread to do the large send; if it didn't, it will likely deadlock as the others do.  On linux, pipe capacity is 65536, and OSX apparently can switch to that too if large writes are made.  But in both cases they can both drop down if too much memory is in use by kernel buffers.

emilio:
Yeah, I think that same-thread sends and recvs shouldn't be allowed. Ideally this would be enforced, but this is hard.

I wonder if Servo uses that a lot... I hope not, but...

antrik:
@vvuk I think the idea was that the "medium_data" test is supposed to be somewhat large, but small enough not to cause fragmentation and thus blocking. (Note that the `linux` backend uses sockets, not pipes -- and in my testing, the buffer size never drops below 104 KiB IIRC, even on serious memory pressure...)

What sizes are we talking about specifically? If it's not way smaller than 64 KiB, we could simply modify the `medium_data` test to stay below the limit...

As for the API, I introduced `platform::OsIpcSender::get_max_fragment_size()` for the benefit of the test cases -- not sure how "official" this should be considered...

antrik:
@vvuk since your current implementation seems to pass the tests just fine, I assume this is no longer an issue?...

vvuk:
Correct, but it only passes the tests since I chose a big enough buffer
size :)

antrik:
@vvuk and you consider that a problem?...

vvuk:
It's potentially surprising behaviour for a consumer of the library.  They can write code like:
```
let (tx, rx) = channel();
tx.send(data);
let data1 = rx.recv().unwrap();
```
[...]

PR #369, opened 2024-10-18
Use get_max_fragment_size for medium_data tests

sunshowers:
Currently, the `medium_data` tests all hang on illumos because (a) we first send a message and then receive it on the same thread, and (b) the tests fill up the pipe buffer before the send is complete. That's because illumos uses a smaller buffer size than other platforms like Linux.

The goal of these tests is to test situations where fragmentation doesn't happen, so use `get_max_fragment_size` which shouldn't fill up the pipe buffer under ordinary circumstances.

(There are other tests which check for fragmentation, so that code path is still covered.)

It would also be nice to move all of the receives to other threads in the future, but this appears to work quite well.

See #101.

### DIFF
--- a/src/platform/test.rs
+++ b/src/platform/test.rs
@@ -108,7 +108,9 @@ fn multisender_transfer() {
 
 #[test]
 fn medium_data() {
-    let data: Vec<u8> = (0..65536).map(|i| (i % 251) as u8).collect();
+    let data: Vec<u8> = (0..get_max_fragment_size())
+        .map(|i| (i % 251) as u8)
+        .collect();
     let data: &[u8] = &data[..];
     let (tx, rx) = platform::channel().unwrap();
     tx.send(data, vec![], vec![]).unwrap();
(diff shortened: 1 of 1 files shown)


jdm (approved):
This looks like a reasonable improvement!

jdm:
Oh, lol—for the macOS implementation the max fragment size is usize::MAX, so that's why we're seeing `capacity overflow` panics in CI.

sunshowers:
> Oh, lol—for the macOS implementation the max fragment size is usize::MAX, so that's why we're seeing `capacity overflow` panics in CI.

Ah I guess that's because the macOS impl uses Mach ports. I'll set a cap on the fragment size used, thanks.

sunshowers:
Updated! Capped it at the previously-used 64KiB value.

sagudev:
Are we sure we need to change these values? I think you need to make sure that fragmentation works on illumos: https://github.com/servo/ipc-channel/blob/4e51f9505bc91cb04409b2eea59ba829b212cf91/src/platform/test.rs#L224-L229

sunshowers:
> Are we sure we need to change these values? I think you need to make sure that fragmentation works on illumos: 

The huge_data tests do exercise that, but good point about those specific tests -- will enable that.

sunshowers:
I've updated the PR summary to indicate what's happening.

sunshowers:
Also enabled the fragment tests on illumos.

sunshowers:
Hi — is there any way I can help get this landed? Thanks!

mrobinson (approved):

mrobinson:
Apologies for the wait and thanks for the fix. This looks good to me!

PR #370, opened 2024-10-18
Accept EINVAL errors in make_socket_lingering

sunshowers:
See included comment for why this is necessary. This is required to make the `server_connect_first` test pass on illumos.


jdm (approved):
This seems reasonable! Thanks for the comment explaining it.

PR #374, opened 2024-11-29
Deprecate old shared memory fd creation

Mikopet:
During the attempt to fix `gaol <> ipc-channel` incompatibilities, I was trying to fix the errors first. the `memfd_create` is fixable on gaol side, but I was not able to fix the `shm_open`.

I cannot really point out the issue. I suspect some tricky bitflip playing around the usernamespace and permissions, but I have no evidence tbh.

---

Anyhow, `memfd_create` is part of the linux kernel from 2014, and also part of FreeBSD from 2021.
I don't see strong reason to keep the older `shm_open` method.

according to [GNU libc manual](https://www.gnu.org/software/libc/manual/html_node/Memory_002dmapped-I_002fO.html#index-shm_005fopen):
- `shm_open`: MT-Safe locale | AS-Unsafe init heap lock | AC-Unsafe lock mem fd
- `memfd_create`: MT-Safe | AS-Safe | AC-Safe fd

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,6 @@ harness = false
 [features]
 default = []
 force-inprocess = []
-memfd = []
 async = ["futures", "futures-test"]
 win32-trace = []
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,13 +23,6 @@ jobs:
           - { target: x86_64-pc-windows-msvc, os: windows-latest }
           - { target: i686-pc-windows-msvc, os: windows-latest }
         features: ["", "force-inprocess", "async"]
-        include:
-          - features: ""
-            platform: { target: x86_64-pc-windows-msvc, os: windows-latest }
-          - features: ""
-            platform: { target: i686-pc-windows-msvc, os: windows-latest }
-          - features: "memfd"
-            platform: { target: x86_64-unknown-linux-gnu, os: ubuntu-latest }
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -26,7 +26,6 @@ harness = false
 [features]
 default = []
 force-inprocess = []
-memfd = []
```
jdm, line 29: Let's leave the feature with a comment saying it's deprecated. This will allow us to put off having to make a semver-major version change.

jdm (approved):

Cargo.toml
Mikopet: Sort of makes sense, but because it's major is `<1`, semver allows to do anything.
It could be minor or patch version dump too.

However, (I just realize) memfd feature is used in the GHA workflow atm, and that also needs some care.
I'd vote for removing memfd feature completely therefore decreasing unnecessary complexity.

what do you think?

sagudev:
BTW for shm_open in gaol, firefox has this in their sandbox code: https://searchfox.org/mozilla-central/source/security/sandbox/linux/Sandbox.cpp#572

Mikopet:
> BTW for shm_open in gaol, firefox has this in their sandbox code: https://searchfox.org/mozilla-central/source/security/sandbox/linux/Sandbox.cpp#572

I tried it out, but same issue. Tbh I cannot fathom what causes it, but it's also not my field of expertise. (evidently not the syscall itself, as audit doesn't report violation. It has to be something in the environment.)

But as it is major incompatibility, I see no reason to support it. (but if somebody can fix it, I'm up for it, sure)

Cargo.toml
jdm: That position is reasonable. I'll merge this as-is.

jdm (approved):

Issue #378, opened 2025-01-21
Inconsistent behaviour of one-shot servers on macOS and Linux


glyn:
While [investigating](https://github.com/glyn/ipc-channel/issues/2) the intended behaviour of one-shot servers, I found an inconsistency between macOS and Linux. I can't tell if the inconsistency is intentional (seems unlikely) or whether the behaviour should be consistent and match the behaviour on macOS or Linux or something else.

The inconsistency is as follows. If multiple client processes connect to a one-shot server and send requests to it, the server can receive all the requests on macOS, but can only receive the requests from one client on Linux.

I think what's happening on macOS is that each client process's connect request accesses the same unidirectional communication channel (described [here](https://developer.apple.com/library/archive/documentation/Darwin/Conceptual/KernelProgramming/Mach/Mach.html#//apple_ref/doc/uid/TP30000905-CH209-TPXREF104)) associated with the Mach port underlying the one-shot server and so the clients' requests all make their way to the channel resulting from the server's call to accept().

I think what's happening on Linux is that each client process's connect request would result in a _distinct_ socket connection returned from the socket accept() underpinning one-shot server's accept(). However, because one-shot server's accept() can only be called once, only the first client connect request is processed and other connect requests wait indefinitely to be accepted.

To make this concrete, the following test passes on macOS, but fails on Linux:
~~~
#[cfg(not(any(
    feature = "force-inprocess",
    target_os = "windows",
    target_os = "android",
    target_os = "ios"
)))]
#[test]
fn cross_process_connecting_multiple_times_to_one_shot_server() {
    let person = ("Arthur Dent".to_owned(), 29);
    let person2 = ("Ford Prefect".to_owned(), 54);
    let (server0, server0_name) = IpcOneShotServer::new().unwrap();
    let child_pid = unsafe {
        fork(|| {
            let tx0 = IpcSender::connect(server0_name.clone()).unwrap();
            tx0.send(person.clone()).unwrap();
        })
    };
    let child_pid2 = unsafe {
        fork(|| {
            let tx0 = IpcSender::connect(server0_name.clone()).unwrap();
            tx0.send(person2.clone()).unwrap();
        })
    };
    let (rx, received_person): (IpcReceiver<Person>, Person) = server0.accept().unwrap();
    let received_person2 = rx.recv().unwrap(); // This panics with "`Err`` value: Disconnected" on Linux
    child_pid.wait();
    child_pid2.wait();
    assert_eq!(received_person, person);
[...]

PR #379, opened 2025-01-22
Document unpredictable behaviour of connect()

glyn:
Ref: https://github.com/servo/ipc-channel/issues/378


jdm (approved):
Thanks for pushing on this!

PR #380, opened 2025-01-23
Deprecate connect() in favour of IpcConnector

glyn:
_This draft pull request prototypes a replacement for connect() with its unpredictable behaviour when called more than once per one-shot server._

Introduce IpcConnector as a means of forcing at most one connect request to a one-shot server.

For a smooth upgrade experience, deprecate new() and connect() in favour of new_with_connector() and its connect() method, respectively.

new() will be deleted in a future release at which point new_with_connector() could be renamed new().

Note that new_with_connector() does not support all the usecases of new(), but should be sufficient for Servo. In particular, it does not support launching a process passing the name of a one-shot server to which the process may then connect.

Ref: https://github.com/servo/ipc-channel/issues/378

### DIFF
--- a/src/ipc.rs
+++ b/src/ipc.rs
@@ -359,7 +359,7 @@ where
     ///
     /// [IpcSender]: struct.IpcSender.html
     /// [IpcOneShotServer]: struct.IpcOneShotServer.html
-    #[deprecated(since="0.20.0", note="please use `new_with_connector` instead")]
+    #[deprecated(since = "0.20.0", note = "please use `new_with_connector` instead")]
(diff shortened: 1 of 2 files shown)


jdm (commented):
Ah, this is an interesting idea! This only works with the fork model though, right? Otherwise there's no way to move the connector into a child process—Servo spawns a process and passes the token as a command line argument.

glyn:
> Ah, this is an interesting idea! This only works with the fork model though, right? Otherwise there's no way to move the connector into a child process—Servo spawns a process and passes the token as a command line argument.

Yes, this only works with fork. Closing, but I think it would be worth capturing the spawn usecase in a testcase in this repo.

PR #381, opened 2025-01-23
Deprecate one-shot server

glyn:
_This draft pull request prototypes a drastic replacement for connect() with its unpredictable behaviour when called more than once per one-shot server._

ipc::channel seems sufficient for Servo's needs.


glyn:
cross_process_embedded_senders_fork_with_channels seems to hang on macOS, so closing to investigate.

glyn:
There was also a panic in cross_process_embedded_senders_fork_with_channels. It seems that forking doesn't play well with the underlying Mach port resources.

Withdrawing this PR.

Issue #382, opened 2025-01-24
Add integration test of a one-shot server client in a spawned process


glyn:
@jdm [pointed out](https://github.com/servo/ipc-channel/pull/380#pullrequestreview-2571752347) that Servo spawns a process, rather than using fork(), to create a child process which acts as a client to a one-shot server in the parent process. It would be good to add an integration test to this repository to cover that usecase.

## Background

A Rust internals thread [Why no `fork()` in `std::process`?](https://internals.rust-lang.org/t/why-no-fork-in-std-process/13770) discusses some of the downsides of fork() and links to the excellent paper [A fork() in the road](https://www.microsoft.com/en-us/research/uploads/prod/2019/04/fork-hotos19.pdf), which describes why spawning a process is much safer than using fork().

The Servo code that spawns a process is [here](https://github.com/servo/servo/blob/e74539404467cc31c03f52a419371258235a1d69/components/constellation/sandboxing.rs#L230). (It sometimes uses servo/gaol as an alternative to spawning a completely new process, so perhaps it would be worth adding a testcase for that usecase too.)

glyn:
To avoid duplication of effort: I'm looking to implement this.

PR #383, opened 2025-01-25
Add integration test

glyn:
Fixes https://github.com/servo/ipc-channel/issues/382


glyn:
Not sure what's wrong with the Darwin checks, but they have been waiting to start for nearly two hours. I wonder if there is an extreme shortage of Darwin runners?

jdm:
Ah, there was a backlog of jobs from each time you force pushed to this PR, and it was only slowly catching up since the pool is quite limited. I've canceled a bunch of the outdated jobs.

jdm (approved):
Thank you!

Issue #310, opened 2023-03-07
Sending an IpcReceiver over a previously sent IpcChannel doesn't work on Windows


zaddach:
I'm trying to set up a secondary channel through which I can transfer Senders/Receivers/SharedMemory. Since the secondary channel has been created in a single process, both its server and client named pipe have the same PID. When the secondary channel's pipe handle gets transferred to the other process (via DuplicateHandle), the PID for its server end isn't updated. When another handle (e.g., from a shared memory blob) is now sent over the connection, the transfer will fail on the other end, as the PID encoded in the OutOfBand message is wrong (and the handle hasn't been duplicated into the receiver's process).

- [The code for getting the server PID is here](https://github.com/servo/ipc-channel/blob/7f432aab1eb77bb2c860881153799087918558e9/src/platform/windows/mod.rs#L1227)
- [The PID check on the receiver side is here](https://github.com/servo/ipc-channel/blob/7f432aab1eb77bb2c860881153799087918558e9/src/platform/windows/mod.rs#L162)
- [Windows' pipefs behavior of setting server/client PID at pipe creation time is documented here](https://googleprojectzero.blogspot.com/2019/09/windows-exploitation-tricks-spoofing.html)

Pseudocode of what I'm trying to do at the sender's side:
```
let sender: IpcSender<Requests> = IpcSender::connect(server_name).unwrap();
let (secondary_sender, secondary_sender_other_end) = channel<SecondaryRequests>().unwrap();
sender.send(Requests::Init(secondary_sender)).unwrap();
let shared_memory = IpcSharedMemory::from_byte(0xba, 1024 * 1024);
// The next command will fail on the receiver side, because the server PID of the secondary OsIpcSender is wrong (still the one of the current process, not the transferred OsIpcReceiver)
secondary_sender.send(SecondaryRequests::Mem(shared_memory)).unwrap();
```

wusyong:
I hit the similar problem when attempting to add multiprocess support for Servo on Windows. It resulted in the same panic in [oob_data](https://github.com/servo/ipc-channel/blob/7f432aab1eb77bb2c860881153799087918558e9/src/platform/windows/mod.rs#L162).

wusyong:
I wonder if we can replace the implementation with [anonymous pipes](https://learn.microsoft.com/en-us/windows/win32/ipc/anonymous-pipe-operations) instead. It states it's safe to duplicate the handle to other process.

zaddach:
[...]

PR #388, opened 2025-03-22
Add mechanism to recover windows pipe handles

tensor-programming:
This PR resolves the following issues: #310 and #277, it also makes it possible to run servo under multiprocess mode on windows (without sandboxing).  

This pull request has undergone multiple iterations, including implementations that utilized anonymous pipes, Advanced Local Procedure Call (ALPC), and a global server responsible for tracking pipe handles across all processes. Each of these approaches had several drawbacks compared to the final implementation presented here.  The final implementation loosely follows a the advice todo from a previous contributor. 

A generic named pipe creation interface function has been introduced to allow the sender and receiver to be instantiated in any order. Prior to this PR, the receiver needed to be created before the sender, which led to various issues when duplicating handles across processes. As a consequence of this change, both endpoints are now marked as duplex, even if bidirectional communication is not utilized. In the future, this functionality could be leveraged to enhance the efficiency of the Windows implementation. Specifically, a check could be performed to determine the existence of a handle, enabling the process to use it as either a sender or receiver as needed.

Additionally, a `recover_handles` method has been added to the `OutOfBandMessage` struct to prevent a panic when an out-of-band message is received from an incorrect handle. If the target process ID (PID) and the current PID differ, this method is invoked to duplicate the target handle into the current process. Since Windows does not allow direct handle transfer between processes, duplication is used as the standard approach. This method applies to channel handles, shared memory handles, and large data receiver handles. It is essential for IPC scenarios where handle transfers are required, ensuring that the current process always has a valid handle.

A set of tests are also included in this PR.  More coverage could be added in the future. 

Here is a showcase of servo running with this PR in Multiprocess mode: 

https://github.com/user-attachments/assets/d2ff20ae-c38c-4df3-8904-7c6d91db0d09

### DIFF
--- a/src/platform/windows/mod.rs
+++ b/src/platform/windows/mod.rs
@@ -134,7 +134,7 @@ pub fn channel() -> Result<(OsIpcSender, OsIpcReceiver), WinError> {
 /// Unify the creation of sender and receiver duplex pipes to allow for either to be spawned first.
 /// Requires the use of a duplex and therefore lets both sides read and write.
 unsafe fn create_duplex(pipe_name: &CString) -> Result<HANDLE, WinError> {
-    if let Ok(handle) = CreateFileA(
+    CreateFileA(
         PCSTR::from_raw(pipe_name.as_ptr() as *const u8),
         FILE_GENERIC_WRITE.0 | FILE_GENERIC_READ.0,
         FILE_SHARE_MODE(0),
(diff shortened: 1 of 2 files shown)


src/platform/windows/mod.rs
```diff
@@ -202,18 +235,7 @@ impl<'data> Message<'data> {
...
-/// That would only work if the intended process a) still exists; b) can be
-/// opened by the receiver with handle dup privileges.  Another approach
-/// could be to use a separate dedicated process intended purely for handle
-/// passing, though that process would need to be global to any processes
-/// amongst which you want to share channels or connect one-shot servers to.
-/// There may be a system process that we could use for this purpose, but
-/// I haven't found one -- and in the system process case, we'd need to ensure
-/// that we don't leak the handles (e.g. dup a handle to the system process,
-/// and then everything dies -- we don't want those resources to be leaked).
+/// These handles are recovered by the `recover_handles` method.
```
wusyong, line 238: So I suppose you chose the approach to recover process handle. Is there any caveat we should aware and probably leave the waning comment in doc comment?

src/platform/windows/mod.rs
```diff
@@ -126,6 +131,37 @@ pub fn channel() -> Result<(OsIpcSender, OsIpcReceiver), WinError> {
...
+            1,
+            // out/in buffer sizes
+            0,
+            PIPE_BUFFER_SIZE as u32,
+            0, // default timeout for WaitNamedPipe (0 == 50ms as default)
+            None,
+        )?;
+
+        Ok(handle)
+    }
```
wusyong, line 162: This can use `or`:

```suggestion
    CreateFileA(
        PCSTR::from_raw(pipe_name.as_ptr() as *const u8),
        FILE_GENERIC_WRITE.0 | FILE_GENERIC_READ.0,
        FILE_SHARE_MODE(0),
        None, // lpSecurityAttributes
        OPEN_EXISTING,
        FILE_ATTRIBUTE_NORMAL,
        None,
    )
    .or(CreateNamedPipeA(
        PCSTR::from_raw(pipe_name.as_ptr() as *const u8),
        PIPE_ACCESS_DUPLEX | FILE_FLAG_OVERLAPPED,
        PIPE_TYPE_BYTE | PIPE_READMODE_BYTE | PIPE_REJECT_REMOTE_CLIENTS,
        // 1 max instance of this pipe
        1,
        // out/in buffer sizes
        0,
        PIPE_BUFFER_SIZE as u32,
        0, // default timeout for WaitNamedPipe (0 == 50ms as default)
        None,
    ))
```

tensor-programming: Of course, is there any advantage between the two besides readability?  Ill just commit it, but I am curious.

src/platform/windows/mod.rs
wusyong: While I believe socket on unix can also achieve similar thing, this is not compatible for current API. I would suggest to open follow-up issue if we want to expand it.

wusyong (approved):

src/platform/windows/mod.rs
tensor-programming: I will change it to use 'or_else' in a follow up PR.  We certainly don't want it to be eagerly evaluated.

tensor-programming: Fixed in #391.

PR #392, opened 2025-05-01
Propagate deserialization errors for IpcSharedMemory

simonwuelker:
This makes `IpcSharedMemory::deserialize` return an error if the deserialized data references an shared memory region region that does not exist.

Part of https://github.com/servo/servo/issues/36792. This doesn't fix the actual issue, but gives users of `ipc-channel` the opportunity to handle the error gracefully. Servo is technically at fault for panicking now, because serialized data being corrupted is not unrealistic.

### DIFF
--- a/src/router.rs
+++ b/src/router.rs
@@ -47,7 +47,10 @@ impl RouterProxy {
         // Router proxy takes both sending ends.
         let (msg_sender, msg_receiver) = crossbeam_channel::unbounded();
         let (wakeup_sender, wakeup_receiver) = ipc::channel().unwrap();
-        thread::spawn(move || Router::new(msg_receiver, wakeup_receiver).run());
+        thread::Builder::new()
+            .name("router-proxy".to_string())
+            .spawn(move || Router::new(msg_receiver, wakeup_receiver).run())
+            .expect("Failed to spawn router proxy thread");
         RouterProxy {
             comm: Mutex::new(RouterProxyComm {
                 msg_sender,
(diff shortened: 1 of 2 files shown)


src/router.rs
```diff
@@ -47,7 +47,10 @@ impl RouterProxy {
         // Router proxy takes both sending ends.
         let (msg_sender, msg_receiver) = crossbeam_channel::unbounded();
         let (wakeup_sender, wakeup_receiver) = ipc::channel().unwrap();
-        thread::spawn(move || Router::new(msg_receiver, wakeup_receiver).run());
+        thread::Builder::new()
+            .name("router-proxy".to_string())
+            .spawn(move || Router::new(msg_receiver, wakeup_receiver).run())
+            .expect("Failed to spawn router proxy thread");
```
simonwuelker, line 53: I gave the router proxy thread a name to make it easier to identify when it panics.

jdm (approved):
Can you update the version to 0.20.0 as well? We haven't published a version with #374 yet.

simonwuelker:
> Can you update the version to 0.20.0 as well? We haven't published a version with https://github.com/servo/ipc-channel/pull/374 yet.

What about https://github.com/servo/ipc-channel/pull/390?

wusyong:
> > Can you update the version to 0.20.0 as well? We haven't published a version with #374 yet.
> 
> What about #390? I suggest we merge this PR first and then #390.

Feel free to open another PR to bump the version as that one is outdated already.
But I suspected the change I made is still relevant

PR #393, opened 2025-05-06
Replace lazy_static! with LazyLock

jschwe:
LazyLock is available since Rust 1.80 in the standard library and allows us to remove the dependency on lazy_static


mrobinson (approved):
Looks good to me as long as we are okay with upgrading our MSRV to 1.8.0.

PR #400, opened 2025-07-21
Remove receivers from Router IpcReceiverSet on channel close.

### LINKED ISSUE #399: Unbounded leak of mach ports when using Router on macOS
Here's a minimized testcase:
```toml
[package]
name = "ipc-fd-leak"
version = "0.1.0"
edition = "2024"

[dependencies]
ipc-channel = "0.20.0"
serde = { version = "1.0.219", features = ["serde_derive"] }
```

```rust
#[derive(serde::Serialize, serde::Deserialize)]
struct CacheMsg(ipc_channel::ipc::IpcSender<Vec<u8>>);

#[derive(serde::Serialize, serde::Deserialize)]
enum Msg {
    Callback(Vec<u8>),
    RequestCallbackViaRouter,
    RequestCallbackViaWorkerThread,
}

fn main() {
    let (cache_sender, cache_receiver) = ipc_channel::ipc::channel::<CacheMsg>().unwrap();
    std::thread::spawn(move || {
        loop {
            match cache_receiver.recv() {
                Ok(msg) => {
                    let _ = msg.0.send(vec![0; 1024]);
                },
                Err(e) => {
                    eprintln!("Error in worker: {e:?}");
                    break;
                },
            }
        }
    });

    let (main_sender, main_receiver) = ipc_channel::ipc::channel::<Msg>().unwrap();
    let main_sender2 = main_sender.clone();
    std::thread::spawn(move || {
        while main_sender2.send(Msg::RequestCallbackViaWorkerThread).is_ok() {
            std::thread::sleep(std::time::Duration::from_millis(10));
        }
    });

    let (router_substitute_sender, router_substitute_receiver) = ipc_channel::ipc::channel::<Vec<u8>>().unwrap();
    let main_sender2 = main_sender.clone();
    std::thread::spawn(move || {
        while let Ok(msg) = router_substitute_receiver.recv() {
            if main_sender2.clone().send(Msg::Callback(msg)).is_err() {
                break;
            }
        }
    });

    while let Ok(msg) = main_receiver.recv() {
        match msg {
            Msg::RequestCallbackViaRouter => {
                let (inner_sender, inner_receiver) = ipc_channel::ipc::channel().unwrap();
                let main_sender = main_sender.clone();
                ipc_channel::router::ROUTER.add_typed_route(
                    inner_receiver,
                    Box::new(move |msg| {
                        let _ = main_sender.send(Msg::Callback(msg.unwrap()));
                    }),
                );
                if let Err(e) = cache_sender.send(CacheMsg(inner_sender)) {
                    eprintln!("Error: {e:?}");
                    break;
                }
            },
            Msg::RequestCallbackViaWorkerThread => {
                if let Err(e) = cache_sender.send(CacheMsg(router_substitute_sender.clone())) {
                    eprintln!("Error: {e:?}");
                    break;
                }
            },
            Msg::Callback(_value) => {},
        }
    }
}
```

The testcase spawns one thread which listens for incoming messages that contain a response channel, then sends a response over the provided channel. The testcase listens on a receiver in the main for one of two messages: a callback, or an instruction to initiate a callback request. Finally, there is a worker thread which sends the callback request instruction to the main thread on a regular interval.

There are two configurations possible in this testcase: RequestCallbackViaRouter and RequestCallbackViaWorkerThread. RequestCallbackViaWorkerThread means that the callback response is sent via a dedicated thread which sends a message to the main thread. RequestCallbackViaRouter uses the ipc_channel ROUTER for the same purpose instead.

When RequestCallbackViaWorkerThread is used, the number of mach ports used by this testcase remains constant. When RequestCallbackViaRouter is used, the number of mach ports increases rapidly without bound, either though the number of channels in the testcase should be the same as the other configuration.

jdm:
Fixes #399.

### DIFF
--- a/src/platform/macos/mod.rs
+++ b/src/platform/macos/mod.rs
@@ -608,7 +608,7 @@ impl OsOpaqueIpcChannel {
 
 pub struct OsIpcReceiverSet {
     port: mach_port_t,
-    ports: Vec<mach_port_t>,
+    ports: Vec<OsIpcReceiver>,
 }
 
 impl OsIpcReceiverSet {
@@ -621,22 +621,28 @@ impl OsIpcReceiverSet {
     }
 
     pub fn add(&mut self, receiver: OsIpcReceiver) -> Result<u64, MachError> {
-        mach_port_move_member(receiver.extract_port(), self.port)?;
-        let receiver_port = receiver.consume_port();
-        self.ports.push(receiver_port);
+        let receiver_port = receiver.extract_port();
+        mach_port_move_member(receiver_port, self.port)?;
+        self.ports.push(receiver);
         Ok(receiver_port as u64)
     }
 
     pub fn select(&mut self) -> Result<Vec<OsIpcSelectionResult>, MachError> {
-        select(self.port, BlockingMode::Blocking).map(|result| vec![result])
+        let result = select(self.port, BlockingMode::Blocking);
+        if let Ok(OsIpcSelectionResult::ChannelClosed(mach_port)) = result {
+            let index = self
+                .ports
+                .iter()
+                .position(|port| port.extract_port() as u64 == mach_port)
+                .expect("Port must be present for error to occur");
+            let _ = self.ports.swap_remove(index);
+        }
+        result.map(|result| vec![result])
     }
 }
 
 impl Drop for OsIpcReceiverSet {
     fn drop(&mut self) {
-        for port in &self.ports {
-            mach_port_mod_release(*port, MACH_PORT_RIGHT_RECEIVE).unwrap();
-        }
         mach_port_mod_release(self.port, MACH_PORT_RIGHT_PORT_SET).unwrap();
     }
 }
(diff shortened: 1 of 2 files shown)


jdm:
This solution isn't right—the unix and inprocess backends already handle removing closed channels from their receivers as part of the select() operation. The macOS backend doesn't, so that's where the fix is needed.

src/platform/macos/mod.rs
```diff
@@ -621,22 +621,28 @@ impl OsIpcReceiverSet {
...
+            let _ = self.ports.swap_remove(index);
+        }
+        result.map(|result| vec![result])
     }
 }
 
 impl Drop for OsIpcReceiverSet {
     fn drop(&mut self) {
-        for port in &self.ports {
-            mach_port_mod_release(*port, MACH_PORT_RIGHT_RECEIVE).unwrap();
```
Taym95, line 638: I guess now OsIpcReceiver owns cleanup?

Taym95 (approved):

src/platform/macos/mod.rs
jdm: Correct.

Issue #402, opened 2025-07-25
Poor test performance


glyn:
With the recent introduction of the `router_flood` test, the performance of `cargo test` has decreased¹:
~~~
cargo test  83.97s user 49.82s system 150% cpu 1:28.82 total
~~~

Prior to this:
~~~
cargo test  17.21s user 1.75s system 132% cpu 14.346 total
~~~

So that's a 315% increase in elapsed time (and a 519% increase in CPU time).

This severly impacts development where small, incremental changes are interspersed with running the tests.

---
¹ Measurements were taken on Linux using the `time` utility.

sagudev:
This is usually solved by gating slow tests behind a feature so they only get run on CI. Maybe we should gate them behind release profile.

Which profile did you used?

glyn:
I didn't knowingly use a profile. Is there a default profile?

BTW the elapsed time impact is _only_ 136% on macOS, which is perhaps how this crept in.

PR #403, opened 2025-07-25
Skip router_flood test for the debug profile

glyn:
The debug profile check was found in:

https://stackoverflow.com/questions/39204908/how-to-check-release-debug-builds-using-cfg-in-rust

Fixes: https://github.com/servo/ipc-channel/issues/402

### DIFF
--- a/src/test.rs
+++ b/src/test.rs
@@ -324,6 +324,7 @@ fn router_simple_global() {
     ROUTER.shutdown();
 }
 
+#[cfg(not(debug_assertions))] // skip the test for the debug profile
 #[test]
 fn router_flood() {
     let router = RouterProxy::new();


glyn:
> We should tests in CI both in release and debug:
> 
> https://github.com/servo/ipc-channel/blob/69d1d4bf8c8dfb6acd70dded299dc03c6c772e14/.github/workflows/main.yml#L41

I agree it's safest to enable the test in CI. I've changed the approach to use a feature to gate the test. So it will no longer depend on debug/release, but on an explicit feature gate. This is also better self-documentation.

Do you think we should also enable the feature for local test runs with the release profile? I think that may be rather surprising behaviour and perhaps best avoided.

glyn:
@sagudev I found this [issue/debate](https://github.com/rust-lang/cargo/issues/2911) about enabling test features by default. tl;dr there are lots of sharp edges and it's probably best avoided.

sagudev:
It think just manually enabling slow test in CI would be enough (I know we have something like this somewhere in servo org).

glyn:
> I think just manually enabling slow test in CI would is enough (I know we have something like this somewhere in servo org).

Happy to switch to another scheme if someone remembers what it is. Meanwhile, thanks for merging.

PR #405, opened 2025-08-01
clippy: GitHub inline annotation

jerensl:
This integrates Clippy errors directly into GitHub inline annotations.

Testing: It has been tested [here](https://github.com/jerensl/ipc-channel/pull/1/files).


jerensl:
@mukilan @mrobinson

etc/ci/clippy-annotation.sh
sagudev, line 1: Maintaining this in every repo is a chore, we should pack this script into github action for easier reuse across servo org and wider rust community.

.github/workflows/main.yml
```diff
@@ -34,8 +34,17 @@ jobs:
       - name: rustfmt
         run: cargo fmt --check
 
+      - name: Make Clippy JSON formatting script executable
+        run: |
+          chmod +x etc/ci/clippy-annotation.sh
```
sagudev, line 39: If you locally do `chmod +x` and commit it everybody should get it with git clone.

jerensl: I didn't know git can track executable bit. Thanks 😅

.github/actions/clippy-annotation/action.yml
```diff
@@ -0,0 +1,27 @@
...
+  using: "composite"
+  steps:
+    - name: Set GitHub Path
+      run: echo "$GITHUB_ACTION_PATH" >> $GITHUB_PATH
+      shell: bash
+      env:
+        GITHUB_ACTION_PATH: ${{ github.action_path }}
+
+    - name: clippy (with annotations)
+      run: cargo clippy --features "$FEATURES" --target "$PLATFORM" --message-format=json | clippy-annotation.sh
```
mukilan, line 23: Looks like we currently don't fail the CI if there are clippy warnings. Should we change the current behaviour with the`-- --deny warnings` argument? If so, this script should also be updated so that the exit code of `cargo clippy` is not swallowed by the `clippy-annotation.sh` script.

.github/workflows/main.yml
```diff
@@ -34,8 +34,16 @@ jobs:
...
+        if: matrix.platform.target == 'x86_64-unknown-linux-gnu' && matrix.features == ''
+        uses: ./.github/actions/clippy-annotation
+        with:
+          features: ${{ matrix.features }}
+          platform-target: ${{ matrix.platform.target }}
+
       - name: clippy
-        run: cargo clippy --features "${{ matrix.features }}" --target ${{ matrix.platform.target }}
+        if: matrix.platform.target != 'x86_64-unknown-linux-gnu' || matrix.features != ''
+        run: cargo clippy --features "${{ matrix.features }}" --target ${{ matrix.platform.target }} --message-format=json
```
mukilan, line 46: Why do we add `--message-format=json` here?

.github/actions/clippy-annotation/action.yml
```diff
@@ -0,0 +1,27 @@
+name: "Clippy Annotation"
+description: "Github inline annotation for clippy"
+inputs:
+  features:
+    description: "Clippy feature to use"
+    required: true
```
mukilan, line 6: If these inputs are `required` we don't need to specify default values.

etc/ci/clippy-annotation.sh
mukilan: > Not yet, but we should make one.

I can create a repo if we are agreed on this plan. Do we have a name in mind? 'github-actions' or just `actions`?

.github/actions/clippy-annotation/action.yml
jerensl: Since we only capture warnings and errors, how about checking the jq output directly and returning an error exit code if any are found as it’s more a simpler approach.

mukilan: I guess that depends on the complexity of the logic to check the `jq` output vs just returning the error code from cargo using the [bash](https://unix.stackexchange.com/questions/14270/get-exit-status-of-process-thats-piped-to-another) syntax. Also, failing the CI job should happen not just when clippy is run with annotations, but also for the other targets/features that currently run `cargo clippy` directly. For those, either we update the command directly or make them use the same action, but with annotations disabled. I think I prefer the latter so we don't have to maintain two `cargo clippy` invocations.

etc/ci/clippy-annotation.sh
mukilan: I've created https://github.com/servo/actions to host the shared CI actions @jerensl Could you create a PR to this repo with the code for the clippy composite action. I'm not sure if the permissions are setup correctly. Let me know on Zulip if you face issues.

.github/actions/clippy-annotation/action.yml
```diff
@@ -0,0 +1,36 @@
...
+    required: false
+    default: "false"
+
+runs:
+  using: "composite"
+  steps:
+    - name: clippy (with annotations)
+      if: inputs.with-annotation == 'true'
+      run: |
+        cargo clippy --features "$FEATURES" --target "$PLATFORM" --message-format=json \
```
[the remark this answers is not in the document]
jerensl: This is an old change. I just waiting for [this PR](https://github.com/servo/actions/pull/1) and then it will like [this examples](https://github.com/jerensl/ipc-channel/pull/2/files#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R38)

mukilan (approved):

PR #406, opened 2025-08-01
fix: Clippy warning with no features on linux platform

jerensl:
Testing: test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.42s

### DIFF
--- a/src/platform/unix/mod.rs
+++ b/src/platform/unix/mod.rs
@@ -589,9 +589,7 @@ impl OsIpcSelectionResult {
         match self {
             OsIpcSelectionResult::DataReceived(id, ipc_message) => (id, ipc_message),
             OsIpcSelectionResult::ChannelClosed(id) => {
-                panic!(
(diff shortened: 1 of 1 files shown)


src/platform/unix/mod.rs
```diff
@@ -863,6 +861,7 @@ impl Deref for OsIpcSharedMemory {
 }
 
 impl OsIpcSharedMemory {
+    /// # Safety
```
sagudev, line 864:
```suggestion
    /// # Safety
    ///
    /// This is safe if there is only one reader/writer on the data.
    /// User can achieve this by not cloning [`IpcSharedMemory`]
    /// and serializing/deserializing only once.
```

from https://docs.rs/ipc-channel/latest/ipc_channel/ipc/struct.IpcSharedMemory.html#method.deref_mut

sagudev (approved):
LGTM - comment

PR #411, opened 2025-08-19
chore: Fix clippy warning on `force-inprocess` feature

jerensl:
This will fix the linting error when running clippy on `force-inprocess` feature

### DIFF
--- a/src/platform/inprocess/mod.rs
+++ b/src/platform/inprocess/mod.rs
@@ -20,7 +20,8 @@ use std::ops::{Deref, RangeFrom};
 use std::slice;
 use std::sync::{Arc, LazyLock, Mutex};
 use std::time::Duration;
-use std::usize;
+use std::ptr::eq;
+use usize;
 use uuid::Uuid;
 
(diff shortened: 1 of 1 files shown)


src/platform/inprocess/mod.rs
```diff
@@ -358,6 +358,7 @@ impl Deref for OsIpcSharedMemory {
 }
 
 impl OsIpcSharedMemory {
+    /// # Safety
```
mrobinson, line 361: You will need to fill out the Safety comment here or ask clippy to ignore the error.

@sagudev I think you created this function. Do you mind writing a small explanation for why it is unsafe.

mrobinson (changes requested):

src/platform/inprocess/mod.rs
sagudev: Same as in https://docs.rs/ipc-channel/latest/ipc_channel/ipc/struct.IpcSharedMemory.html#method.deref_mut

mrobinson (approved):

PR #413, opened 2025-08-27
router: ensure the router proxy loop is broken out of on shutdown

gterzian:
The current shutdown does a break, but that only breaks out of the inner event-handling loop, so the thread doesn't actually shut down. This replaces it with a return, and also joins on the thread handle for good measure. Includes unrelated changes from an fmt. 

Testing: added a unit test.

### DIFF
--- a/src/router.rs
+++ b/src/router.rs
@@ -15,12 +15,14 @@
 
 use std::collections::HashMap;
 use std::sync::{LazyLock, Mutex};
-use std::thread;
+use std::thread::{self, JoinHandle};
 
-use crate::ipc::OpaqueIpcReceiver;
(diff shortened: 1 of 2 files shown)


mrobinson (approved):
Nice catch.

src/router.rs
```diff
@@ -117,6 +120,11 @@ impl RouterProxy {
                 ack_receiver.recv().unwrap();
             })
             .unwrap();
+        comm.handle
+            .take()
+            .expect("Should have a join handle at shutdown")
+            .join()
+            .expect("Failed to join on the router proxy thread");
```
Taym95, line 127: nit: we can avoid panicking here:

```
if let Err(err) = handle.join() {
    error!("Router thread panicked during shutdown: {:?}", err);
}
```

Taym95 (approved):

PR #420, opened 2025-09-27
Install target with rustup.

jdm:
This should fix the following x86 CI error:
```
Error: error[E0463]: can't find crate for `core`
  |
  = note: the `i686-pc-windows-msvc` target may not be installed
  = help: consider downloading the target with `rustup target add i686-pc-windows-msvc`
```


wusyong (approved):

jschwe (commented):
You can add targets in the install section with the dtolnay action a couple of lines above.

PR #421, opened 2025-10-17
Cleaned up memory dance in ipc::send and ipc::to.

Narfinger:
After this discussion (https://servo.zulipchat.com/#narrow/channel/263398-general/topic/IPC-Channel.20question/with/545510321)
and my own understanding the memory dance done does not seem to be necessary.
Additionally, we can replace the uses of std::mem with methods on the RefCell which makes the code cleaner and easier to understand. I also added some comments that future people are not confused why we handle memory in a thing that should serialize.

This might increase performance by a bit as my tests show 1-24% increase in the bench with most in the single digit. I noticed regressions
that seem to be just outliers as the surrounding benchmarks are fine.
On constraint devices this might be more of an increase in performance (15-42%) but these devices are very noisy.

Testing:
The included tests still work and a whole run of servospeedometer with this version of ipc-channel still worked. All of these tests were only done on Linux but it should be independent.

### DIFF
--- a/src/ipc.rs
+++ b/src/ipc.rs
@@ -20,7 +20,6 @@ use std::error::Error as StdError;
 use std::fmt::{self, Debug, Formatter};
 use std::io;
 use std::marker::PhantomData;
-use std::mem;
 use std::ops::Deref;
 use std::time::Duration;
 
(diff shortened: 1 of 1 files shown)


jdm (approved):
This is definitely easier to follow now!

PR #426, opened 2025-11-20
Expose a way to manually set BOOTSTRAP_PREFIX

ChristianBelloni:
To be able to use this library in a sandboxed environment the prefix has to adhere to a specific format, [here's a link to a relevant SO discussion](https://stackoverflow.com/questions/9889186/cfmessageport-and-sandboxing)

I have no idea what would be the ideal approach to correctly expose the bootstrap prefix, in this pr it's basically slapped on in lib.rs, if this is a use case that you'd like to support let me know any changes needed to merge

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -40,3 +40,5 @@ pub mod router;
 mod test;
 
 pub use bincode::{Error, ErrorKind};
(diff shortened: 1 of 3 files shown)


jdm (approved):
This seems like a reasonable way to support this.

PR #430, opened 2025-12-19
Switch from fnv to rustc_hash

Narfinger:
Fnv hash is slower than rustc_hash especially for integer keys.


Cargo.toml
sagudev: We should remove fnv here.

sagudev (approved):

PR #431, opened 2026-01-02
Notice callback drop timing in `add_route`

Legend-Master:
I have tried to look up if I could listen to channel disconnects on a `IpcReceiver` when I'm using the `ROUTER` and only found out that the `callback` is dropped on receiver's channel disconnects so I could move a drop guard in to the handler after looking into the actual implementation, so I thought it might be good to document this

### DIFF
--- a/src/router.rs
+++ b/src/router.rs
@@ -69,7 +69,7 @@ impl RouterProxy {
     /// Add a new (receiver, callback) pair to the router, and send a wakeup message
     /// to the router.
     ///
-    /// The `callback` is dropped when `receiver`'s channel disconnects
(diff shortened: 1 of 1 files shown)


src/router.rs
```diff
@@ -83,6 +85,8 @@ impl RouterProxy {
 
     /// Add a new `(receiver, callback)` pair to the router, and send a wakeup message
     /// to the router.
+    ///
+    /// The `callback` is dropped when `receiver`'s channel disconnects
```
glyn:
```suggestion
    /// The `callback` is dropped when `receiver`'s channel disconnects.
```

PR #436, opened 2026-01-13
Added test for drop of sender

Narfinger:
This adds a test for dropping the sender when a message is not received yet.
Currently this behavior was unspecified and untested. This gives us now an expected behavior.

### DIFF
--- a/src/test.rs
+++ b/src/test.rs
@@ -135,7 +135,8 @@ fn simple() {
 }
 
 #[test]
-/// We first want to have all the data that is an empty channel and only then the disconnect.
+// Messages sent before a channel is disconnected should be received before
+// recv returns the disconnected error.
 fn disconnect_non_empty_channel() {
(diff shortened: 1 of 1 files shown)


src/test.rs
```diff
@@ -134,6 +134,15 @@ fn simple() {
     }
 }
 
+#[test]
+fn drop_equal_not_disconnect_with_msg() {
+    let person = ("Patrick Walton".to_owned(), 29);
+    let (tx, rx) = ipc::channel().unwrap();
+    tx.send(person.clone()).unwrap();
+    drop(tx);
+    assert_eq!(rx.recv().unwrap(), ("Patrick Walton".to_owned(), 29));
```
sagudev: Why else do we clone person (no pun intended) if not for:
```suggestion
    assert_eq!(rx.recv().unwrap(), person);
```

sagudev (approved):
Specification by tests is better then no specification at all.

